PR #3, opened 2024-03-22
Swerve creation -> master

YoniKiriaty:
kind of copied from ctre but more like the code we have. its been changed a bit but you can see whats the same and whats not by looking [here](https://github.com/dirtbikerxz/BaseTalonFXSwerve/blob/main/src/main/java/frc/robot/subsystems/Swerve.java)


NoamRosenberg08:
Where is the akit impelmentatiob?

PR #28, opened 2024-04-07
yoav - Gversion (5 files)

YHerm:


src/main/java/frc/utils/LoggerUtils.java
```diff
@@ -1,5 +1,6 @@
 package frc.utils;
 
+import frc.buildconstants.BuildConstants;
```
YoniKiriaty, line 3: build constants is not a thing.
did you forget to do `git add .`?

YoniKiriaty (changes requested):

YoniKiriaty (approved):

PR #30, opened 2024-04-08
yoav - Auto Commit (1 file)

YHerm:


build.gradle
```diff
@@ -60,15 +63,47 @@ repositories {
...
+                }
+
+                println("commited to branch -> '$branch'")
+                println("commited message -> '$commitMessage'")
+            } else {
+                println("skipping -> not a branch to auto commit")
+            }
+        } else {
+            println("skipping -> not a deploy or simulate")
+        }
```
YoniKiriaty, line 102: instead of doing else at the end, use the inverse conditions put it first to make it cleaner.
```suggestion
        if (!taskNames.any({ it.toLowerCase().contains("deploy") }) && !taskNames.any({ it.toLowerCase().contains("simulate") })) {
            println("skipping -> not a deploy or simulate")
            return
        }

        def eventPrefix = "event"
        def fieldPrefix = "field"
        def checkPrefix = "check"
        def branch = 'git branch --show-current'.execute().text.trim()
        def commitMessage = "autoCommit - updated at '${new Date().toString()}'"

        if (!branch.startsWith(eventPrefix) && !branch.startsWith(fieldPrefix) && !branch.startsWith(checkPrefix)) {
            println("skipping -> not a branch to auto commit")
            return
        }
        exec {
            workingDir(projectDir)
            executable('git')
            args 'commit', '-am', commitMessage
            ignoreExitValue = true
        }

        println("commited to branch -> '$branch'")
        println("commited message -> '$commitMessage'")
```

YoniKiriaty (changes requested):

build.gradle
YHerm: Why , it is making the if more complicated to understand

PR #45, opened 2024-04-13
yoav - System Identification 

YHerm:


src/main/java/frc/utils/calibration/sysid/SysIdObject.java
```diff
@@ -0,0 +1,110 @@
...
+
+    public Command getSysIdDynamic(SysIdRoutine.Direction direction) {
+        final Command command = sysIdRoutine.dynamic(direction);
+        return addRequirementsAndReturnCommand(command);
+    }
+
+    private Command addRequirementsAndReturnCommand(Command sysIdCommand) {
+        sysIdCommand.addRequirements(usedSubSystem);
+        return isCTRE ? getCTRECommand(sysIdCommand) : sysIdCommand;
+    }
```
YoniKiriaty, line 100: shouldn't this be two seperate functions?
`addRequierments` and `getAprropriateCommand`?

YHerm: dont see a reason to split it, i mean i want the ctre command the be alone in function, so add one more function to do the check doesnt feel necessery, but i can do that

YoniKiriaty: a function should only have one purpose, split it

YHerm: no, when you split this func you give a option to get a command without add requiremenrts, also this function have one purpose and that to return appropriate command even if i split it this function will do 2 things

YoniKiriaty: image

YoniKiriaty (commented):
I actually think you should just rename the original function to something like the getApropriateCommand since you just return the good command, and you should delete the getApropriateCommand that is already there (idk how to use github from phone)

YoniKiriaty (approved):

src/main/java/frc/utils/calibration/sysid/SysIdObject.java
```diff
@@ -0,0 +1,134 @@
...
+import edu.wpi.first.wpilibj2.command.sysid.SysIdRoutine;
+import frc.utils.GBSubsystem;
+import org.littletonrobotics.junction.Logger;
+
+import java.util.function.Consumer;
+
+import static edu.wpi.first.units.Units.Seconds;
+import static edu.wpi.first.units.Units.Volts;
+
+public class SysIdObject {
```
YoniKiriaty, line 18: I don't think `SysIdObject` is a good name, consider changing to something like `SysIdCalibrator`

PR #85, opened 2024-04-27
yoav - Roborio Utils

YHerm:


src/main/java/frc/utils/RoborioUtils.java
```diff
@@ -0,0 +1,29 @@
...
+public class RoborioUtils {
+
+    private static double lastTime;
+
+    private static double currentTime;
+
+    public static void updateCurrentCycleTime() {
+        lastTime = currentTime;
+        currentTime = Timer.getFPGATimestamp();
+    }
```
YoniKiriaty, line 15: shouldn't you use this function in a periodic?
if you think it should be another CR that's fine.

YHerm: wow that was fast

YoniKiriaty (dismissed):

YoniKiriaty (approved):

YoniKiriaty (approved):

PR #96, opened 2024-04-28
yoav - Pigeon2Wrapper

YHerm:


src/main/java/frc/utils/devicewrappers/Pigeon2Wrapper.java
YoniKiriaty, line 1: why not GBPigeon2?

src/main/java/frc/utils/devicewrappers/TalonFXWrapper.java
YoniKiriaty, line 1: why not GBTalonFXPro?

src/main/java/frc/utils/devicewrappers/Pigeon2Wrapper.java
YHerm: I dont like do gb names. also thats a wrapper we didnt add something new that is our. Wrapper is better name i think. If it is important i can change

YHerm: You can use the original, this wrapper meant to help with working with the object and this is actually a wrapper. If you a suggestion to a better name i would love to hear.

YoniKiriaty (approved):

PR #97, opened 2024-04-28
yoav - Dashboard Command

YHerm:


src/main/java/frc/utils/commands/DashboardCommand.java
```diff
@@ -0,0 +1,26 @@
...
+            String widgetName,
+            Consumer<Double> methodToRun,
+            GBSubsystem... subsystems
+    ) {
+        super(
+                () -> SmartDashboard.putNumber(widgetName, startingValue),
+                () -> methodToRun.accept(SmartDashboard.getNumber(widgetName, startingValue)),
+                subsystems
+        );
+    }
```
YoniKiriaty, line 24: consider adding a constructor where there is no starting value

YoniKiriaty (approved):

PR #108, opened 2024-04-29
yoav -  Wheel Radius Characterization

YHerm:


src/main/java/frc/utils/calibration/swervecalibration/CharacterizationDirection.java
```diff
@@ -0,0 +1,17 @@
...
+
+    private final int directionSign;
+
+    CharacterizationDirection(int directionSign) {
+        this.directionSign = directionSign;
+    }
+
+    public int getDirectionSign() {
+        return directionSign;
+    }
```
YoniKiriaty, line 16: why not make the direction sign public and get rid of the getter?
you cannot change the direction sign because it is final so i see no reason not to

YoniKiriaty (dismissed):

PR #111, opened 2024-04-30
hillel - adding autonomous selector (3 files)

hillelv:


YHerm (commented):
Look good. Consider changing the selector to be static class (your choice this is not very important).

YHerm (commented):
Try to do that you dont need to call buildSelector from other classes and it initialize itself

hillelv:
> Try to do that you dont need to call buildSelector from other classes and it initialize itself

You have to initialize it when the robot initializes so you can choose the auto

YHerm (commented):
Can't you create it as static method in class, it doesnt seem that the autobuilder require you to run new Robot

YoniKiriaty (dismissed):

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -27,6 +27,7 @@ public class RobotContainer {
     public RobotContainer() {
         // Configure the trigger bindings
         configureBindings();
+        AutonomousSelector.buildSelector();
```
YHerm, line 30: you are not suppose to create objects in this block, this block is for applying commands as bindings and maybe to schedule some more non binding commands.
Consider to move it to other place or make it a variable in this class like a subsystem (robot init, or other place if you have)

src/main/java/frc/utils/AutonomousSelector.java
```diff
@@ -0,0 +1,31 @@
...
+
+    public static Command getChosenValue() {
+        if (selector != null) {
+            Logger.recordMetadata("autonomous chosen", selector.getSelected().getName());
+            return selector.getSelected();
+        }
+        else {
+            Logger.recordMetadata("autonomous chosen", "None");
+            return new InstantCommand();
+        }
```
YHerm, line 29: I know i suggested static but consider change it again to not static, also it would remove this if else block.

src/main/java/frc/utils/AutonomousSelector.java
```diff
@@ -0,0 +1,31 @@
...
+import edu.wpi.first.wpilibj.shuffleboard.Shuffleboard;
+import edu.wpi.first.wpilibj.shuffleboard.ShuffleboardTab;
+import edu.wpi.first.wpilibj.smartdashboard.SendableChooser;
+import edu.wpi.first.wpilibj2.command.Command;
+import edu.wpi.first.wpilibj2.command.InstantCommand;
+import org.littletonrobotics.junction.Logger;
+
+public class AutonomousSelector {
+
+    private static SendableChooser<Command> selector;
```
YHerm, line 13: If it would not change consider to make it final

src/main/java/frc/utils/AutonomousSelector.java
hillelv: now that the selector is final there's no need for the block.
I think the static implementation is cleaner because we only need one, and I see the selector as a utility to access its data

hillelv: I can move the selector to a final in robot container and make the class an actual util if you want

YHerm (approved):
The new design looks very nice!

PR #112, opened 2024-04-30
Working Trigon Swerve (for NOAM Rosenberg)

YHerm:
In progress, getting discretize and coupling fix, designs fixes, and improve with aim assist


YoniKiriaty (approved):
nice job

src/main/java/frc/robot/constants/FieldConstants.java
```diff
@@ -1,9 +1,21 @@
...
 
+    public static final MirrorableTranslation3d SPEAKER = new MirrorableTranslation3d(
+            0.23,
+            FIELD_WIDTH_METERS - 2.55,
+            2.045,
+            true
+    );
+
+    public static final MirrorableRotation2d ANGLE_TO_AMP = MirrorableRotation2d.fromDegrees(90, false);
+
```
levyishai, line 20: Why is this false for mirroring? You'd want to face the other way when you're on the red side.

levyishai (changes requested):

src/main/java/frc/robot/constants/FieldConstants.java
YHerm: its more code that runs, and mirror does nothing so i thought to give up on setting true

PR #119, opened 2024-05-04
for hilel - Fix Auto Selector

YHerm:
For the next time, please run the code before putting "passed-all-tests" label!!!


src/main/java/frc/robot/Robot.java
```diff
@@ -33,6 +34,8 @@ public void robotInit() {
         // autonomous chooser on the dashboard.
         initializeLogger();
         robotContainer = new RobotContainer();
+
+        AutonomousSelector.addSelectorToShuffleboard();
```
YHerm, line 38: you will change desgin so it's not important, but you use the selector in robot container so a good approach is to create (add to dashboard) things before usage. I mean above and not under the new RobotConatiner()

hillelv: This needs to be below so that the auto builder configures in swerve init happens before building the selector

PR #177, opened 2024-06-08
yoav - Swerve Depend Constants

YHerm:
Where to save these const? (variable at swerve or at swerve constants)
What name to give them instead of depend?


src/main/java/frc/robot/subsystems/swerve/swervedependconstants/SwerveDependConstants.java
hillelv, line 1: maybe call it SwerveTypeDependantConstants

PR #195, opened 2024-06-26
yoav: Paths To Logger

YHerm:
image

I made the simulation path independent of the computer system.

There are 2 options to do it:
The one in the picture (but then the line is long).
Or what I did in this PR.
What do you prefer?

### DIFF
--- a/src/main/java/frc/utils/loggerutils/LoggerUtils.java
+++ b/src/main/java/frc/utils/loggerutils/LoggerUtils.java
@@ -8,6 +8,8 @@ import org.littletonrobotics.junction.networktables.NT4Publisher;
 import org.littletonrobotics.junction.wpilog.WPILOGReader;
 import org.littletonrobotics.junction.wpilog.WPILOGWriter;
 
+import java.nio.file.Path;
+
 
 public class LoggerUtils {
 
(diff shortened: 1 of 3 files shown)


src/main/java/frc/utils/loggerutils/LoggerUtils.java
```diff
@@ -47,7 +47,7 @@ public static void startReplayLogger() {
     }
 
     private static void startNonReplayLogger(LogSaveSpot logSaveSpot) {
-        setLoggingPath(logSaveSpot.savePath);
+        setLoggingPath(logSaveSpot.savePath.toString());
```
YoniKiriaty, line 50: Why does setLoggingPath get a String as a parameter and not a Path?

YoniKiriaty (changes requested):

YoniKiriaty:
> What do you prefer?

I prefer the way you did it in the CR

src/main/java/frc/utils/loggerutils/LoggerUtils.java
YHerm: Because if i give a path i would need to do toString twice or save var with the string, and its duplicate code

YoniKiriaty: But the name suggests it gets a path. If you don't want to turn every path from a String to a Path I see no reason to make some paths different

YHerm: I am not sure what i prefer, but the functions in gets string so what do you think?

YoniKiriaty: I think that if everything gets a string and not a path there is no reason to change the Constants into paths

YHerm: First of all it's a better approach to handle paths as paths, like u use enums for stuff and not just 1 2 3 4.
The second thing is the line in the image is very long so i need it to be shorter

YoniKiriaty: I know but I think if you are using Paths you should go all the way and not just put a .to string() in the middle of the process.
Could you imagine if I put angles in a rotation2d but then I converted it into a double in the middle of my code and used it like that in my functions?

YHerm: look on the code again

YHerm: i changed it

bossyshrimp68 (approved):
nice

PR #224, opened 2024-07-12
Subsystems Log Path -> master

NoamRosenberg08:


tnrz135 (dismissed):

src/main/java/frc/utils/GBSubsystem.java
```diff
@@ -6,6 +6,12 @@
...
 
+    private final String loggingPath;
+
+    public GBSubsystem(String loggingPath) {
+        this.loggingPath = "Subsystems/" + loggingPath;
+    }
+
     @Override
     public final void periodic() {
         Logger.recordOutput(getLogPath() + "Current Command", getCurrentCommandName());
```
YoniKiriaty, line 17: Why are you doing getLogPath if loggingPath is already a variable in this class?

YHerm: We can do public final but its ugly

YoniKiriaty: I don't want you to switch to public final, but you can just use loggingPath

YoniKiriaty: I meant only in this line change it

src/main/java/frc/utils/GBSubsystem.java
```diff
@@ -6,6 +6,12 @@
 
 public abstract class GBSubsystem extends SubsystemBase {
 
+    private final String logPath;
+
+    public GBSubsystem(String logPath) {
+        this.logPath = "Subsystems/" + logPath;
```
yonnbro, line 12: why remove from constants?

yonnbro (changes requested):

bossyshrimp68 (approved):

PR #233, opened 2024-07-13
Logged Table Boolean -> master

YHerm:


src/main/java/frc/utils/dashboard/LoggedTableBoolean.java
```diff
@@ -0,0 +1,44 @@
...
+    public LoggedTableBoolean(String table, String key) {
+        this(table, key, false);
+    }
+
+    public LoggedTableBoolean(String table, String key, boolean defaultValue) {
+        this.key = key;
+        this.value = defaultValue;
+        this.booleanTable = NetworkTableInstance.getDefault().getTable(table);
+
+        set(defaultValue);
```
YoniKiriaty, line 24: This causes the issue of the first input staying, find a way to fix that or don't set the value

YoniKiriaty (changes requested):

src/main/java/frc/utils/dashboard/LoggedTableBoolean.java
YHerm: U literally did that

PR #236, opened 2024-07-13
Path planner util -> pp-dev

hillelv:


src/main/java/frc/utils/pathplannerutils/PathPlannerUtils.java
```diff
@@ -0,0 +1,94 @@
...
+import java.util.function.Supplier;
+
+public class PathPlannerUtils {
+
+    public static void startPathPlanner() {
+        makePathPlannerCompatibleWithAdvantageKit();
+        scheduleWarmup();
+    }
+
+    private static void makePathPlannerCompatibleWithAdvantageKit() {
```
hillelv, line 33:
```suggestion
    private static void setAdvantageKitCompatiblePathfinder() {
```

YHerm (commented):
Test it on robot, if you need swerve or other stuff open test branch and write the tester and take swerve in it.

src/main/java/frc/utils/auto/LocalADStarAK.java
```diff
@@ -0,0 +1,156 @@
...
+     *
+     * @return True if a new path is available
+     */
+    @Override
+    public boolean isNewPathAvailable() {
+        if (!Logger.hasReplaySource()) {
+            io.updateIsNewPathAvailable();
+        }
+
+        Logger.processInputs("LocalADStarAK", io);
```
YoniKiriaty, line 34: maybe save the log path as a constant. i think we have a constants file for these but maybe they are in another CR

src/main/java/frc/utils/auto/PathPlannerUtils.java
```diff
@@ -0,0 +1,94 @@
...
+        NamedCommands.registerCommand(commandName, command);
+    }
+
+    public static void setDynamicObstacles(List<Pair<Translation2d, Translation2d>> obstacles, Pose2d currentRobotPose) {
+        Pathfinding.setDynamicObstacles(obstacles, currentRobotPose.getTranslation());
+    }
+
+    public static void removeAllDynamicObstacles(Pose2d currentRobotPose) {
+        setDynamicObstacles(List.of(), currentRobotPose);
+    }
```
YoniKiriaty, line 71: consider adding an `addDynamicObstacle`

src/main/java/frc/utils/auto/PathPlannerUtils.java
```diff
@@ -0,0 +1,94 @@
...
+import java.util.function.Supplier;
+
+public class PathPlannerUtils {
+
+    public static void startPathPlanner() {
+        makePathPlannerCompatibleWithAdvantageKit();
+        scheduleWarmup();
+    }
+
+    private static void makePathPlannerCompatibleWithAdvantageKit() {
```
hillelv, line 33:
```suggestion
    private static void setAdvantageKitCompatiblePathfinder() {
```

src/main/java/frc/utils/auto/PathPlannerUtils.java
```diff
@@ -0,0 +1,94 @@
...
+
+import java.util.List;
+import java.util.Optional;
+import java.util.function.BooleanSupplier;
+import java.util.function.Consumer;
+import java.util.function.Supplier;
+
+public class PathPlannerUtils {
+
+    public static void startPathPlanner() {
```
hillelv, line 28:
```suggestion
    public static void startPathfinder() {
```

src/main/java/frc/utils/auto/LocalADStarAK.java
```diff
@@ -0,0 +1,156 @@
...
+import org.littletonrobotics.junction.Logger;
+import org.littletonrobotics.junction.inputs.LoggableInputs;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+public class LocalADStarAK implements Pathfinder {
+
+    private final ADStarIO io = new ADStarIO();
```
bossyshrimp68, line 21: rename

src/main/java/frc/utils/auto/PathPlannerUtils.java
```diff
@@ -0,0 +1,94 @@
...
+            Consumer<Pose2d> resetPose,
+            Supplier<ChassisSpeeds> robotRelativeSpeedsSupplier,
+            Consumer<ChassisSpeeds> robotRelativeOutput,
+            HolonomicPathFollowerConfig config,
+            BooleanSupplier shouldFlipPath,
+            Subsystem driveSubsystem
+    ) {
+        AutoBuilder.configureHolonomic(
+                poseSupplier,
+                resetPose,//todo - maybe cancel, if vision very accurate solving wrong robot placement
```
bossyshrimp68, line 52: ask amir

bossyshrimp68 (changes requested):

src/main/java/frc/utils/auto/LocalADStarAK.java
hillelv: The constants file is for file directories and not log paths, for log paths we mostly use the class's constants file. I don't want to create a whole constants file for this

src/main/java/frc/utils/auto/PathPlannerUtils.java
YHerm: delete the comment, if the people don't want to reset they will give supplier that do nothing

bossyshrimp68 (changes requested):

bossyshrimp68 (approved):
just fix the space line

src/main/java/frc/utils/auto/PathPlannerUtils.java
```diff
@@ -0,0 +1,97 @@
...
+
+	public static void removeAllDynamicObstacles(Pose2d currentRobotPose) {
+		setDynamicObstacles(List.of(), currentRobotPose);
+	}
+
+	public static void setTargetRotationOverride(Supplier<Optional<Rotation2d>> overrider) {
+		PPHolonomicDriveController.setRotationTargetOverride(overrider);
+	}
+
+	public static Command createOnTheFlyPathCommand(Pose2d currentBluePose, Pose2d targetPose, PathConstraints constraints) {
```
YHerm, line 90: What is blue pose?

src/main/java/frc/utils/auto/PathPlannerUtils.java
```diff
@@ -0,0 +1,97 @@
...
+		scheduleWarmup();
+	}
+
+	private static void setPathfinder(Pathfinder pathfinder) {
+		Pathfinding.setPathfinder(pathfinder);
+	}
+
+	private static void scheduleWarmup() {
+		PathfindingCommand.warmupCommand().schedule();
+	}
```
YHerm, line 44: can be public

src/main/java/frc/utils/auto/PathPlannerUtils.java
```diff
@@ -0,0 +1,97 @@
...
+
+	public static void configurePathPlanner(
```
YHerm, line 46:
```suggestion
	public static void configPathPlanner(
```
its a convention in swerve

src/main/java/frc/utils/auto/PathPlannerUtils.java
YHerm, line 90:
```suggestion
	public static Command createPathOnFly(Pose2d currentBluePose, Pose2d targetPose, PathConstraints constraints) {
```

src/main/java/frc/utils/auto/PathPlannerUtils.java
```diff
@@ -0,0 +1,97 @@
...
+
+	public static void setDynamicObstacles(List<Pair<Translation2d, Translation2d>> obstacles, Pose2d currentRobotPose) {
+		dynamicObstacles = obstacles;
+		Pathfinding.setDynamicObstacles(obstacles, currentRobotPose.getTranslation());
+	}
+
+	public static void addDynamicObstacles(List<Pair<Translation2d, Translation2d>> obstacles, Pose2d currentRobotPose) {
+		List<Pair<Translation2d, Translation2d>> allObstacles = new ArrayList<>();
+		allObstacles.addAll(dynamicObstacles);
+		allObstacles.addAll(obstacles);
```
YHerm, line 78: you can just add it to the dynamicObstacles

src/main/java/frc/utils/auto/PathPlannerUtils.java
```diff
@@ -0,0 +1,97 @@
...
+		setDynamicObstacles(allObstacles, currentRobotPose);
+	}
+
+	public static void removeAllDynamicObstacles(Pose2d currentRobotPose) {
+		setDynamicObstacles(List.of(), currentRobotPose);
+	}
```
YHerm, line 84: can you find out if robot pose is really needed here, or can be no params func.
also it looks weird that it needs robot pose if it have a supplier of it, why is that?

src/main/java/frc/utils/auto/PathPlannerUtils.java
```diff
@@ -0,0 +1,97 @@
...
+
+	private static void scheduleWarmup() {
+		PathfindingCommand.warmupCommand().schedule();
+	}
+
+	public static void configurePathPlanner(
+		Supplier<Pose2d> poseSupplier,
+		Consumer<Pose2d> resetPose,
+		Supplier<ChassisSpeeds> robotRelativeSpeedsSupplier,
+		Consumer<ChassisSpeeds> robotRelativeOutput,
```
YHerm, line 50: i know that its pp naming but it is every weird name

src/main/java/frc/utils/auto/PathPlannerUtils.java
hillelv: the current pose is used only if the obstacles are blocking the path for recalculation, so we can just put new Pose2d() instead as it won't happen.
as for the supplier, the auto builder gives it to the PathfindingCommand which gives the pose to the pathfinder via setter

src/main/java/frc/utils/auto/PathPlannerUtils.java
hillelv: do you want me to change it to currentRobotPose or currentBlueAlliancePose?

src/main/java/frc/utils/auto/PathPlannerUtils.java
```diff
@@ -0,0 +1,97 @@
...
+			shouldFlipPath,
+			driveSubsystem
+		);
+	}
+
+	public static void registerCommand(String commandName, Command command) {
+		NamedCommands.registerCommand(commandName, command);
+	}
+
+	public static void setDynamicObstacles(List<Pair<Translation2d, Translation2d>> obstacles, Pose2d currentRobotPose) {
```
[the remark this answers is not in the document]
hillelv: Only if the last path went through an obstacle 
In remove you can "lie" because the path won't be recalculated, but in set and add you need the real pose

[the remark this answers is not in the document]
hillelv: No obstacles so nothing blocking the path so the path doesn't need to be recalculated 

That is the only use of the pose in the function 
Screenshot_2024-09-05-16-28-00-490_com android chrome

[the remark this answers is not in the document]
YHerm: weird because if i am in the middle of a path and remove obstacles it "clear" my way and can do quicker path.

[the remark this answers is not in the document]
hillelv: Maybe there's a way to force recalculate, but then we definitely need the real pose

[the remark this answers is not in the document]
YHerm: i dont know if it can harm to not give pose, and i think we will use this func in functions that gets pose (for the path) so maybe its better to get the pose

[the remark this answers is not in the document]
hillelv: I'm sure the pathfinder set obstacles function won't set the current pose in remove, but maybe the pathfinding thread recalculates the path automatically. I need to read more to be sure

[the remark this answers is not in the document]
hillelv: Ok so the thread only recalculates the path if goal or starting pose is set. 
If you want to force recalculation I can set starting position in remove (it could make a faster path but it will increase code complexity)

src/main/java/frc/utils/auto/PathPlannerUtils.java
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	public static void setRotationTargetOverride(Supplier<Optional<Rotation2d>> overrider) {
+		PPHolonomicDriveController.setRotationTargetOverride(overrider);
+	}
+
+	public static Command createPathOnTheFly(Pose2d currentRobotPose, Pose2d targetPose, PathConstraints constraints) {
+		List<Translation2d> bezierPoints = PathPlannerPath.bezierFromPoses(currentRobotPose, targetPose);
+		PathPlannerPath path = new PathPlannerPath(bezierPoints, constraints, new GoalEndState(0, targetPose.getRotation()));
+		path.preventFlipping = true;
```
bossyshrimp68, line 93: magic boolean

bossyshrimp68 (changes requested):

YoniKiriaty (approved):

src/main/java/frc/utils/auto/PathPlannerUtils.java
```diff
@@ -0,0 +1,104 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import java.util.function.BooleanSupplier;
+import java.util.function.Consumer;
+import java.util.function.Supplier;
+
+public class PathPlannerUtils {
+
+	private static final boolean DEFAULT_PREVENT_PATH_FLIPPING = true;
```
bossyshrimp68, line 31: why the default

src/main/java/frc/utils/auto/PathPlannerUtils.java
```diff
@@ -0,0 +1,104 @@
...
+
+	public static void removeAllDynamicObstacles(Pose2d currentRobotPose) {
+		setDynamicObstacles(List.of(), currentRobotPose);
+	}
+
+	public static void setRotationTargetOverride(Supplier<Optional<Rotation2d>> overrider) {
+		PPHolonomicDriveController.setRotationTargetOverride(overrider);
+	}
+
+	public static Command
```
bossyshrimp68, line 92: disable fornater here

bossyshrimp68 (changes requested):

src/main/java/frc/utils/auto/PathPlannerUtils.java
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	// @formatter:off
+	public static Command createPathOnTheFly(Pose2d currentRobotPose, Pose2d targetPose, PathConstraints constraints, boolean preventFlipping) {
+	// @formatter:on
+		List<Translation2d> bezierPoints = PathPlannerPath.bezierFromPoses(currentRobotPose, targetPose);
+		PathPlannerPath path = new PathPlannerPath(bezierPoints, constraints, new GoalEndState(0, targetPose.getRotation()));
+		path.preventFlipping = preventFlipping;
+		return AutoBuilder.followPath(path);
+	}
```
YHerm, line 99: or current/targetRobotPose or current/targetPose, pick one

src/main/java/frc/utils/auto/PathPlannerUtils.java
```diff
@@ -0,0 +1,105 @@
...
+	public static void setRotationTargetOverride(Supplier<Optional<Rotation2d>> overrider) {
+		PPHolonomicDriveController.setRotationTargetOverride(overrider);
+	}
+
+	// @formatter:off
+	public static Command createPathOnTheFly(Pose2d currentRobotPose, Pose2d targetPose, PathConstraints constraints, boolean preventFlipping) {
+	// @formatter:on
+		List<Translation2d> bezierPoints = PathPlannerPath.bezierFromPoses(currentRobotPose, targetPose);
+		PathPlannerPath path = new PathPlannerPath(bezierPoints, constraints, new GoalEndState(0, targetPose.getRotation()));
+		path.preventFlipping = preventFlipping;
```
YHerm, line 97: i think u always want to prevent when using on fly

YHerm (changes requested):

bossyshrimp68 (approved):

src/main/java/frc/utils/auto/PathPlannerUtils.java
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	public static void setRotationTargetOverride(Supplier<Optional<Rotation2d>> overrider) {
+		PPHolonomicDriveController.setRotationTargetOverride(overrider);
+	}
+
+	public static Command createPathOnTheFly(Pose2d currentPose, Pose2d targetPose, PathConstraints constraints) {
+		List<Translation2d> bezierPoints = PathPlannerPath.bezierFromPoses(currentPose, targetPose);
+		PathPlannerPath path = new PathPlannerPath(bezierPoints, constraints, new GoalEndState(0, targetPose.getRotation()));
+		path.preventFlipping = PREVENT_PATH_FLIPPING;
```
YHerm, line 95: I dont think this should be const

YHerm (approved):

PR #237, opened 2024-07-13
Swerve Containers -> swerve constants

YHerm:


bossyshrimp68 (approved):
love it!

src/main/java/frc/robot/Robot.java
```diff
@@ -16,7 +17,11 @@ public class Robot {
     public static final RobotTypeUtils.RobotType ROBOT_TYPE = RobotTypeUtils.determineRobotType(RobotTypeUtils.RobotType.REAL);
 
 
-    public static final Swerve swerve = new Swerve();
+    public static final Swerve swerve = new Swerve(
+            SwerveFactory.createConstants(),
+            SwerveFactory.createModules(),
+            SwerveFactory.createGyro()
+    );
```
YoniKiriaty, line 24: consider maybe adding another swerve constructor with these paramaters

YHerm: Like record? If yes I don’t see a reason, like it’s just one more class

PR #240, opened 2024-07-15
Log table bool -> master

YHerm:


src/main/java/frc/utils/dashboard/LoggedTableBoolean.java
```diff
@@ -0,0 +1,44 @@
...
+    public LoggedTableBoolean(String table, String key) {
+        this(table, key, false);
+    }
+
+    public LoggedTableBoolean(String table, String key, boolean defaultValue) {
+        this.key = key;
+        this.value = defaultValue;
+        this.booleanTable = NetworkTableInstance.getDefault().getTable(table);
+
+        set(defaultValue);
```
YHerm, line 24: This, function makes the first time the value become true keep true instead of getting false until triggering the entry again.

tnrz135 (approved):

bossyshrimp68 (approved):

tnrz135 (approved):

PR #254, opened 2024-07-24
Configs Applier -> master

greenblitz4590:


bossyshrimp68 (commented):
re-request review when name is fixed

src/main/java/frc/utils/ctre/PhoenixProUtils.java
```diff
@@ -1,11 +1,23 @@
...
+import java.util.function.Supplier;
+
 public class PhoenixProUtils {
 
     public static <T> StatusSignal<T> getRefreshedSignal(boolean refresh, StatusSignal<T> signal) {
         return refresh ? signal.refresh() : signal;
     }
 
-}
+    public static boolean checkAndRetry(Supplier<StatusCode> statusCodeSupplier, int tries){
```
bossyshrimp68, line 14:
```suggestion
    public static boolean checkAndRetry(Supplier<StatusCode> statusCodeSupplier, int numberOfTries){
```

bossyshrimp68 (changes requested):

bossyshrimp68 (approved):
good job!!

src/main/java/frc/utils/ctre/PhoenixProUtils.java
```diff
@@ -1,11 +1,23 @@
...
 
+import java.util.function.Supplier;
+
 public class PhoenixProUtils {
 
     public static <T> StatusSignal<T> getRefreshedSignal(boolean refresh, StatusSignal<T> signal) {
         return refresh ? signal.refresh() : signal;
     }
 
+    public static boolean checkAndRetry(Supplier<StatusCode> statusCodeSupplier, int numberOfTries) {
```
tnrz135, line 14: This function name suggests it has  2 responsibilities, as it has "and" in its name, which is also not very clean and breaks the Single Responsibility Principle. But actually this function does only one thing and that is the retries. Consider a better name for the function.

tnrz135 (changes requested):

bossyshrimp68 (approved):

PR #280, opened 2024-08-05
Battery Message -> master

YHerm:


src/main/java/frc/utils/battery/BatteryLimiter.java
```diff
@@ -3,18 +3,44 @@
...
 import frc.utils.DriverStationUtils;
+import frc.utils.dashboard.LoggedTableBoolean;
 import org.littletonrobotics.junction.Logger;
 
+import java.nio.file.Path;
+
 class BatteryLimiter extends Command {
 
     private static final int NUMBER_OF_SAMPLES_TAKEN_IN_AVERAGE = 50;
+    private static final double MESSAGE_STARTUP_TIME = 0.2; // 2 python cycles
```
tnrz135, line 19: I would prefer to split this to 2 constants of python cycle time and number of cycles to wait.

Also, what do you mean by python cycle time?

src/main/java/frc/utils/battery/BatteryLimiter.java
```diff
@@ -34,10 +60,16 @@ public void execute() {
...
-            if (GlobalConstants.ENABLE_BATTERY_LIMITER && !DriverStationUtils.isMatch()) {
+            showBatteryMessage();
+            if (GlobalConstants.ENABLE_BATTERY_LIMITER && !DriverStationUtils.isMatch() && passedMessageStartUpTime()) {
                 throw new java.lang.RuntimeException("BATTERY IS LOW");
             }
         }
+        else {
+            if (isBatteryLow.get()) {
+                isBatteryLow.set(false);
+            }
```
tnrz135, line 71: CTRL Alt L

src/main/java/frc/utils/battery/BatteryLimiter.java
tnrz135, line 1: I get the idea with the image, but it's not good enough. While testing the robot during the season, you will look on the robot with a hand on the keyboard. You can very easily miss the warning or not be able to stop in time.

tnrz135 (changes requested):

src/main/java/frc/utils/battery/BatteryLimiter.java
YHerm: The loop in python runs every 100 ms

src/main/java/frc/utils/battery/BatteryLimiter.java
YHerm: Must of the times the code will throw too, but if not I believe you will notice pop up window, if not I don’t now what also we can do

src/main/java/frc/utils/battery/BatteryLimiter.java
YHerm: And we check connection to robot and then check if we should show the message, so we need to wait 2 cycles for it to surely be able to show message before stopping

src/main/java/frc/utils/battery/BatteryLimiter.java
```diff
@@ -34,10 +61,14 @@ public void execute() {
         double currentAverageVoltage = voltageFilter.calculate(BatteryUtils.getCurrentVoltage());
         if (currentAverageVoltage <= BatteryUtils.MIN_VOLTAGE) {
             reportLowBattery();
-            if (GlobalConstants.ENABLE_BATTERY_LIMITER && !DriverStationUtils.isMatch()) {
+            showBatteryMessage();
+            if (GlobalConstants.ENABLE_BATTERY_LIMITER && !DriverStationUtils.isMatch() && passedMessageStartUpTime()) {
```
bossyshrimp68, line 65: you check if its a match after you do the popup... you should still state that its low but not with a massive Noam

src/main/python/BatteryMessage.py
```diff
@@ -0,0 +1,73 @@
+import sys
+import time
+import tkinter as tk
```
YoniKiriaty, line 3: consider not using `as`

src/main/java/frc/utils/battery/BatteryLimiter.java
```diff
@@ -34,9 +61,14 @@ public void execute() {
 		double currentAverageVoltage = voltageFilter.calculate(BatteryUtils.getCurrentVoltage());
 		if (currentAverageVoltage <= BatteryUtils.MIN_VOLTAGE) {
 			reportLowBattery();
-			if (GlobalConstants.ENABLE_BATTERY_LIMITER && !DriverStationUtils.isMatch()) {
-				throw new java.lang.RuntimeException("BATTERY IS LOW");
+			if (!DriverStationUtils.isMatch()) {
+				showBatteryMessage();
+				if (GlobalConstants.ENABLE_BATTERY_LIMITER && passedMessageStartUpTime()) {
+					throw new java.lang.RuntimeException("BATTERY IS LOW");
```
bossyshrimp68, line 67: import

bossyshrimp68 (changes requested):

bossyshrimp68 (approved):

bossyshrimp68 (approved):

src/main/java/frc/utils/battery/BatteryLimiter.java
tnrz135: My issue is that the code will throw...

tnrz135 (changes requested):

tnrz135 (dismissed):

src/main/python/BatteryMessage.py
```diff
@@ -0,0 +1,71 @@
...
+from tkinter import PhotoImage
+from NetworkTableManager import NetworkTableClient
+
+IMAGE_PATH = "../../../noam-battery-message.png"
+WINDOW_NAME = "Battery Message"
+
+CLIENT_NAME = "BatteryMessage"
+LOW_BATTERY_TOPIC_NAME = "BatteryMessage/LowBattery"
+
+TIME_BETWEEN_MESSAGES_SECONDS = 4
```
YHerm, line 13: Change to 60

YHerm: will be last change, needed for testing

PR #320, opened 2024-08-25
looper-> master

GilStein1:
a wrapper for a for loop


aaa-png (dismissed):

src/main/java/frc/utils/GBLooper.java
```diff
@@ -0,0 +1,13 @@
+package frc.utils;
+
+import java.util.function.Consumer;
+
+public class GBLooper {
```
YHerm, line 5: why GB?

YHerm, line 5: call it a util name and not object

src/main/java/frc/utils/LoopingUtils.java
```diff
@@ -0,0 +1,13 @@
...
+import java.util.function.Consumer;
+
+public class LoopingUtils {
+
+	public static void loop(Consumer<Integer> loop, int iterations) {
+		for (int index = 0; index < iterations; index++) {
+			loop.accept(index);
+		}
+	}
+
```
NoamRosenberg08, line 12: loop (Consumer<Integer>... loops )

NoamRosenberg08 (approved):

src/main/java/frc/utils/LoopingUtils.java
NoamRosenberg08: nvm

PR #342, opened 2024-09-08
Input Signal -> core

YHerm:


src/main/java/frc/robot/hardware/signal/InputSignal.java
```diff
@@ -0,0 +1,70 @@
...
+			this.toAngle = toAngle;
+		}
+
+		protected void setNewValues(double[] newValues) {
+			setNewValues(Arrays.stream(newValues).mapToObj(toAngle).toArray(Rotation2d[]::new));
+		}
+
+		protected void setNewValues(double newValue) {
+			setNewValues(toAngle.apply(newValue));
+		}
```
YoniKiriaty, line 37: If i have something that returns a rotation 2d, I won't need to use the DoubleFunction and the functions `setNewValues`
. But according to  your code I will need to provide a toAngle even though i won't use it

this can be fixed with a default constructor that doesn't require a function

src/main/java/frc/robot/hardware/signal/InputSignal.java
```diff
@@ -0,0 +1,70 @@
...
+import org.littletonrobotics.junction.LogTable;
+import org.littletonrobotics.junction.inputs.LoggableInputs;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.function.DoubleFunction;
+
+
+public abstract class InputSignal<T> implements LoggableInputs {
+
```
Sh-Pe, line 13: Before laying out instances of how the class functions, it makes much more sense to describe _what_ the class do and how it functions. Hence, it makes more sense to put the abstract class at the end of the class, or in a  different file.

src/main/java/frc/robot/hardware/signal/InputSignal.java
```diff
@@ -0,0 +1,70 @@
...
+		protected AngleSignal(DoubleFunction<Rotation2d> toAngle) {
+			super(new Rotation2d());
+			this.toAngle = toAngle;
+		}
+
+		protected void setNewValues(double[] newValues) {
+			setNewValues(Arrays.stream(newValues).mapToObj(toAngle).toArray(Rotation2d[]::new));
+		}
+
```
Sh-Pe, line 34: Maybe add a way to change the angle in-place. It makes more sense in the code, depends on the usecase, and is more efficient when it can be used.

Sh-Pe (changes requested):
I could haven't understand the code at the first time I read it -> it's bad (jk). 

Also, תקפיד להיות מאוד פורמלי, תציין כל שטות של "לפי עקרון ההפרדה" "מכלל בטא" "מתחשיב למדא" "לפי הגדרת הכלה" וכיו"ב.

PR #388, opened 2024-09-29
Poseestimator fixed

Sh-Pe:
small fixes and new filters

### DIFF
--- a/src/main/java/frc/utils/FieldStartingPositions.java
+++ b/src/main/java/frc/utils/FieldStartingPositions.java
@@ -4,6 +4,8 @@ import edu.wpi.first.math.geometry.Pose2d;
 
 public enum FieldStartingPositions {
 
+    // ! should have data from the actual field
+    //TODO: add those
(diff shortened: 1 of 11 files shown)


Sh-Pe:
can't simulate

src/main/java/frc/utils/FieldStartingPositions.java
```diff
@@ -0,0 +1,26 @@
...
+
+	private final Pose2d startingPose;
+
+	FieldStartingPositions(Pose2d startingPose) {
+		this.startingPose = startingPose;
+	}
+
+	public Pose2d getStartingPose() {
+		return this.startingPose;
+	}
```
YoniKiriaty, line 24: yoav: we can start in every pose in the little wing

aaa-png (approved):

src/main/java/frc/utils/FieldStartingPositions.java
Sh-Pe: wdym by 'a little wing'

PR #400, opened 2024-10-03
Cutpoint -> core-swerve

drorta:


src/main/java/frc/robot/subsystems/swerve/swervestatehelpers/AimAssistMath.java
```diff
@@ -67,4 +67,17 @@ public static double applyMagnitudeCompensation(Rotation2d velocityPerSecond, do
 			/ (magnitude + SwerveConstants.AIM_ASSIST_MAGNITUDE_FACTOR);
 	}
 
+	public static double[] getClosetCutPoint(double x, double y, double R){
```
Nitay4, line 70:
```suggestion
	public static double[] getClosestCutPoint(double x, double y, double R){
```

Nitay4 (changes requested):

src/main/java/frc/robot/subsystems/swerve/swervestatehelpers/AimAssistMath.java
```diff
@@ -67,4 +67,17 @@ public static double applyMagnitudeCompensation(Rotation2d velocityPerSecond, do
 			/ (magnitude + SwerveConstants.AIM_ASSIST_MAGNITUDE_FACTOR);
 	}
 
+	public static Translation2d getClosestCutPoint(Translation2d robot, Translation2d speaker, double ShootingZoneRadius){
```
hillelv, line 70:
```suggestion
	public static Translation2d getClosestCutPoint(Translation2d robot, Translation2d target, double ShootingZoneRadius){
```

src/main/java/frc/robot/subsystems/swerve/swervestatehelpers/AimAssistMath.java
```diff
@@ -67,4 +67,17 @@ public static double applyMagnitudeCompensation(Rotation2d velocityPerSecond, do
...
+		double m = robot.getY() / robot.getX();
```
hillelv, line 71: שיפוע (אבל יענו בלעז)

hillelv (changes requested):

src/main/java/frc/robot/subsystems/swerve/swervestatehelpers/AimAssistMath.java
YHerm, line 70:
```suggestion
	public static Translation2d getClosetPointInCircle(Translation2d robot, Translation2d centerOfCircle, double radius){
```

src/main/java/frc/robot/subsystems/swerve/swervestatehelpers/AimAssistMath.java
```diff
@@ -67,4 +67,17 @@ public static double applyMagnitudeCompensation(Rotation2d velocityPerSecond, do
...
+		double cutPointX = Math.sqrt(
+				Math.pow(ShootingZoneRadius,2)
+				/ (Math.pow(m,2) + 1)
+		);
+		if (robot.getX() < 0) {
+			cutPointX *= -1;
+		}
```
YHerm, line 78: will not work (imagine if the speaker of middle of field)

YHerm (changes requested):

src/main/java/frc/robot/subsystems/swerve/swervestatehelpers/AimAssistMath.java
drorta: this calculation is normalizing it to thinking the speaker is it 0,0 and then just adding the center x,y. why wont it work?

YHerm: why <0

YHerm: you act like the robot always have a bigger x than the target

drorta: The result x of the calculation will always be positive, so you check if the origin x was negative because if it was u want to keep it that way

src/main/java/frc/robot/subsystems/swerve/swervestatehelpers/AimAssistMath.java
```diff
@@ -67,4 +67,17 @@ public static double applyMagnitudeCompensation(Rotation2d velocityPerSecond, do
 			/ (magnitude + SwerveConstants.AIM_ASSIST_MAGNITUDE_FACTOR);
 	}
 
+	public static Translation2d getClosetPointInCircle(Translation2d robot, Translation2d centerOfCircle, double radius){
+		double slope = robot.getY() / robot.getX();
+		double cutPointX = Math.sqrt(
+				Math.pow(radius,2)
+				/ (Math.pow(slope,2) + 1)
+		);
```
YHerm, line 75: one line

src/main/java/frc/robot/subsystems/swerve/swervestatehelpers/AimAssistMath.java
```diff
@@ -67,4 +67,17 @@ public static double applyMagnitudeCompensation(Rotation2d velocityPerSecond, do
...
+		if (robot.getX() < 0) {
+			cutPointX *= -1;
+		}
```
YHerm, line 78: use signnum

src/main/java/frc/robot/subsystems/swerve/swervestatehelpers/AimAssistMath.java
```diff
@@ -67,4 +67,17 @@ public static double applyMagnitudeCompensation(Rotation2d velocityPerSecond, do
 			/ (magnitude + SwerveConstants.AIM_ASSIST_MAGNITUDE_FACTOR);
 	}
 
+	public static Translation2d getClosetPointInCircle(Translation2d robot, Translation2d centerOfCircle, double radius){
```
YHerm, line 70: u say in circle but its on circle (add reference to what happen when in circle)

drorta: what?

YHerm: what happens if i am in shooting zone (in circle)?
i will get point further from my current pose which is not good

Sh-Pe (commented):
I think using wpilib's function for plus and minus from translation (instead of doing this manually) would make the code more clear. Your choice. Ans run spotless.

src/main/java/frc/robot/subsystems/swerve/swervestatehelpers/AimAssistMath.java
```diff
@@ -67,4 +67,12 @@ public static double applyMagnitudeCompensation(Rotation2d velocityPerSecond, do
...
+		double slope = (robot.getY()-centerOfCircle.getY()) / (robot.getX()-centerOfCircle.getX());
+		double cutPointX = Math.sqrt(Math.pow(radius,2) / (Math.pow(slope,2) + 1));
```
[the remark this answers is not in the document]
hillelv: power of -2 is not sqrt, pow of 0.5 is sqrt
either way don't do pow instead of sqrt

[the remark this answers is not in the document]
Sh-Pe: right my bad, it should be `-0.5`
but anyway I think you should factor out the $r^2$

PR #402, opened 2024-10-03
Linear Filters ⭆ PoseEstimator

Sh-Pe:
currently a draft. Will be merged to core, iff poseestimator is merged, and will not be merged to poseestimator.


Sh-Pe:
can't simulate

PR #410, opened 2024-10-10
Time Utils -> core

YHerm:


bossyshrimp68 (dismissed):

drorta (dismissed):

src/main/java/frc/utils/time/TimeUtils.java
```diff
@@ -0,0 +1,47 @@
...
+import edu.wpi.first.hal.HALUtil;
+import frc.utils.Conversions;
+import frc.utils.alerts.Alert;
+import frc.utils.alerts.AlertManager;
+import frc.utils.alerts.PeriodicAlert;
+import org.littletonrobotics.junction.Logger;
+
+public class TimeUtils {
+
+	public static final double DEFAULT_CYCLE_TIME_SECONDS = 0.02;
```
Sh-Pe, line 12: Why isn’t the roborio's cycle time a global constant? It's being used in other files too.

src/main/java/frc/utils/time/TimeUtils.java
```diff
@@ -0,0 +1,47 @@
...
+
+	private static double lastCycleTimeSeconds = 0;
+	private static double newCycleTimeSeconds = 0;
+
+	static {
+		AlertManager.addAlert(
```
Sh-Pe, line 18: Consider adding an alert for timestamps that are too old (3s>) or that are newer than the current time. This could catch errors in the timestamp's calculation.

Sh-Pe (commented):
I guess now the poseestimator isn’t cringing anymore

src/main/java/frc/utils/time/TimeUtils.java
YHerm: Where?

src/main/java/frc/utils/time/TimeUtils.java
YHerm: Because it related to here, in global there should be constants that don’t have their own class and being used in non related files

src/main/java/frc/utils/time/TimeUtils.java
Sh-Pe: nvm

Sh-Pe (approved):

hillelv:
Alerts use timer too

Yotam7 (approved):

GilStein1 (approved):
Very good👌

hillelv (approved):

drorta (approved):

PR #426, opened 2024-10-26
Test up to date -> core

YHerm:


drorta (approved):
אהבתי

.scripts/post-checkout
hillelv, line 1: dev branch and test branch are two different concepts. test is a branch with testers, and dev is a branch with all approved and tested parts of a complex feature

PR #427, opened 2024-10-28
Comp Syncopa

greenblitz4590:


src/main/java/frc/utils/Conversions.java
```diff
@@ -126,6 +126,16 @@ public static double microSecondsToSeconds(double microSeconds) {
...
 
+	/**
+	 * Converts a time from milliseconds to seconds
+	 *
+	 * @param milliSeconds the milliseconds
+	 * @return the seconds
+	 */
+	public static double milliSecondsToSeconds(double milliSeconds) {
+		return milliSeconds / 1e3;
+	}
```
greenblitz4590, line 137: its also here

src/main/java/frc/utils/joysticks/SmartJoystick.java
```diff
@@ -8,7 +8,7 @@
 
 public class SmartJoystick {
 
-	private static final double DEADZONE = 0.07;
+	private static final double DEADZONE = 0.11;
```
greenblitz4590, line 11: yeah prob this too

PR #436, opened 2024-10-31
Combine angle/double request (Smooch AI Smooch) -> core

YHerm:


src/main/java/frc/robot/hardware/motor/phoenix6/TalonFXMotor.java
```diff
@@ -50,17 +50,11 @@ public void setPower(double power) {
...
-			motor.setControl(((Phoenix6DoubleRequest) request).getControlRequest());
-		}
-	}
-
-	@Override
-	public void applyAngleRequest(IRequest<Rotation2d> request) {
-		if (request instanceof Phoenix6AngleRequest) {
-			motor.setControl(((Phoenix6AngleRequest) request).getControlRequest());
+	public void applyRequest(IRequest request) {
+		if (request instanceof Phoenix6Request<?>) {
```
emilyyyylime, line 54:
```suggestion
		if (request instanceof Phoenix6Request<?> req) {
```

src/main/java/frc/robot/hardware/motor/phoenix6/TalonFXMotor.java
```diff
@@ -50,17 +50,11 @@ public void setPower(double power) {
...
+			motor.setControl(((Phoenix6Request<?>) request).getControlRequest());
+		} else {
+			new Alert(Alert.AlertType.WARNING, getLogPath() + "got invalid type of request").report();
```
emilyyyylime, line 57: Can you print the received request as well?

src/main/java/frc/robot/hardware/request/cansparkmax/SparkMaxRequest.java
```diff
@@ -0,0 +1,60 @@
...
+
+	SparkMaxRequest(T setPoint, CANSparkBase.ControlType controlType, int pidSlot, Function<T, Double> setPointToDoubleConverter) {
+		this(setPoint, controlType, pidSlot, CANSparkMAX -> 0.0, setPointToDoubleConverter);
+	}
+
+	@Override
+	public SparkMaxRequest<T> withSetPoint(T setPoint) {
+		this.setPoint = setPoint;
+		return this;
+	}
```
emilyyyylime, line 38: Given a function name like `withSetPoint` I wouldn't expect `this` to be modified, but a new request object to be created. I do however see how the designed use case makes it so this makes no difference. I'll leave this at your discretion.

src/main/java/frc/robot/hardware/request/cansparkmax/SparkMaxRequestBuilder.java
```diff
@@ -0,0 +1,29 @@
+package frc.robot.hardware.request.cansparkmax;
+
+import com.revrobotics.CANSparkBase;
+import edu.wpi.first.math.geometry.Rotation2d;
+
+import java.util.function.Function;
+
+public class SparkMaxRequestBuilder {
+
+	//@formatter:off
```
emilyyyylime, line 10: What does the formatter do? If you manually need to turn it on and off I consider that a bug in the formatter config.

src/main/java/frc/robot/hardware/request/cansparkmax/SparkMaxRequestBuilder.java
```diff
@@ -0,0 +1,29 @@
...
+    public SparkMaxRequest<Rotation2d> generate(
```
emilyyyylime, line 11: If it's a "builder" wouldn't these be called `build()`? With the current name this seems like a `Generator`. Actually I don't see why not just add these as static functions of `SparkMaxRequest`, I think that's cleaner

emilyyyylime: To clarify, you'd write `SparkMaxRequest.with(args...)` (or whatever name you choose).
If you want to be extra explicit with named constructors, you can have `.revolutionsPerSecond(Rotation2d)`, `.voltage(Double)`, and `.amperage(Double)` (keeping the control type as more of an implementation detail)

emilyyyylime (dismissed):

src/main/java/frc/robot/hardware/request/cansparkmax/SparkMaxRequestBuilder.java
YHerm: The func name line down, and then one line from then

src/main/java/frc/robot/hardware/request/cansparkmax/SparkMaxRequestBuilder.java
YHerm: It cannot have multiple setpoints, so how would it work

emilyyyylime: those would be static methods, not instance methods. But indeed the name `with` might be confusing then. Probably `.of` is fine or `construct` or `newVoltage`/`newRPS`/...

YHerm: u r breaking SRP mi friendo. It does not really matter but i prefer it as diffrent builder

src/main/java/frc/robot/hardware/request/cansparkmax/SparkMaxRequestBuilder.java
emilyyyylime: This can be fixed by changing `/codestyle.xml:124`
```diff
-        <setting id="org.eclipse.jdt.core.formatter.alignment_for_method_declaration" value="16"/>
+        <setting id="org.eclipse.jdt.core.formatter.alignment_for_method_declaration" value="0"/>
```
Wanna sneak that into this PR or should I open a separate one?

src/main/java/frc/robot/hardware/request/cansparkmax/SparkMaxRequestBuilder.java
emilyyyylime: yeah that's fine. Just make sure the method and class names match up

emilyyyylime (approved):

src/main/java/frc/robot/hardware/request/phoenix6/Phoenix6RequestBuilder.java
```diff
@@ -0,0 +1,27 @@
...
+import com.ctre.phoenix6.controls.PositionVoltage;
+import com.ctre.phoenix6.controls.TorqueCurrentFOC;
+import com.ctre.phoenix6.controls.VelocityVoltage;
+import com.ctre.phoenix6.controls.VoltageOut;
+import edu.wpi.first.math.geometry.Rotation2d;
+
+public class Phoenix6RequestBuilder {
+
+	public static Phoenix6Request<Rotation2d> build(PositionVoltage positionVoltage) {
+		return new Phoenix6Request<>(new Rotation2d(), positionVoltage, setPoint -> positionVoltage.withPosition(setPoint.getRotations()));
```
hillelv, line 12: why do you create a default set point in the builder here and get it as parameter in spark max builder

src/main/java/frc/robot/hardware/request/cansparkmax/SparkMaxRequestBuilder.java
YHerm: another pr
there are a lot of places of that in code need to remove format off from them

emilyyyylime: on it

hillelv (dismissed):

src/main/java/frc/robot/hardware/motor/Controllable.java
inigomontoyaik, line 1: herman u r kof❤️🥰🤩

inigomontoyaik: did i du ryt¿

YoniKiriaty (approved):

drorta (approved):

hillelv (approved):

bossyshrimp68 (approved):

src/main/java/frc/robot/hardware/request/cansparkmax/SparkMaxRequest.java
```diff
@@ -0,0 +1,61 @@
...
```
Sh-Pe, line 38: `withSetPoint` is more useful, but I would implement `setSetPoint` too.

src/main/java/frc/robot/hardware/request/cansparkmax/SparkMaxRequest.java
```diff
@@ -0,0 +1,61 @@
...
+		return controlType;
+	}
+
+	public int getPidSlot() {
+		return pidSlot;
+	}
+
+	public double getFeedforwardCalculation() {
+		return feedforwardCalculator.apply(setPoint);
+	}
```
Sh-Pe, line 59:
```suggestion
	public double getFeedforwardGain() {
		return feedforwardCalculator.apply(setPoint);
	}
```


ff is a function that returns a gain in the power, I think it's better to call it with the name of the output.

src/main/java/frc/robot/hardware/request/phoenix6/Phoenix6Request.java
```diff
@@ -0,0 +1,36 @@
...
+
+	Phoenix6Request(T defaultSetPoint, ControlRequest controlRequest, Consumer<T> withSetPoint) {
+		this.setPoint = defaultSetPoint;
+		this.controlRequest = controlRequest;
+		this.withSetPoint = withSetPoint;
+	}
+
+	@Override
+	public Phoenix6Request<T> withSetPoint(T setPoint) {
+		withSetPoint.accept(setPoint);
```
Sh-Pe, line 22: an example for a use case in which `setSetPoint` is better

PR #439, opened 2024-11-01
Cycle count and Average Cycle time -> core

drorta:


src/main/java/frc/utils/time/TimeUtils.java
```diff
@@ -11,6 +11,8 @@ public class TimeUtils {
 
 	public static final double DEFAULT_CYCLE_TIME_SECONDS = 0.02;
 
+	private static int CYCLES = 0;
```
YHerm, line 14: bro is doing changble constants

src/main/java/frc/utils/time/TimeUtils.java
```diff
@@ -34,6 +36,9 @@ public static void updateCycleTime() {
 	private static void logStatus() {
 		Logger.recordOutput(TimeConstants.LOG_PATH + "CycleTimeSeconds", getCurrentCycleTimeSeconds());
 		Logger.recordOutput(TimeConstants.LOG_PATH + "CurrentTimeSeconds", getCurrentTimeSeconds());
+		CYCLES++;
```
YHerm: i would put the cycles counter in robot manager, its not relevant to time

src/main/java/frc/robot/RobotManager.java
```diff
@@ -28,6 +29,8 @@ public class RobotManager extends LoggedRobot {
 
 	private Robot robot;
 
+	public static int cycles;
```
YHerm, line 32: public static above private members

src/main/java/frc/robot/RobotManager.java
```diff
@@ -70,6 +73,8 @@ public void teleopInit() {
 	@Override
 	public void robotPeriodic() {
 		TimeUtils.updateCycleTime(); // Better to be first
+		cycles++;
+		Logger.recordOutput("Cycles", cycles);
```
YHerm, line 77: put this before the time update

src/main/java/frc/robot/RobotManager.java
YHerm, line 32:
```suggestion
	public static int roborioCyclesCounter/roborioCycles;
```

src/main/java/frc/robot/RobotManager.java
YHerm, line 77: rename too

YHerm (approved):
labels

hillelv (dismissed):

YHerm (dismissed):

src/main/java/frc/robot/RobotManager.java
```diff
@@ -69,6 +72,8 @@ public void teleopInit() {
 
 	@Override
 	public void robotPeriodic() {
+		roborioCycles++;
+		Logger.recordOutput("roborioCycles", roborioCycles);
```
YHerm, line 76:
```suggestion
		Logger.recordOutput("RoborioCycles", roborioCycles);
```

src/main/java/frc/robot/RobotManager.java
```diff
@@ -69,6 +72,8 @@ public void teleopInit() {
 
 	@Override
 	public void robotPeriodic() {
+		roborioCycles++;
+		Logger.recordOutput("RoborioCycles", roborioCycles);
 		TimeUtils.updateCycleTime(); // Better to be first
```
Yotam7, line 77: why is there a comment how this line should be first, but it isn't first.
Either remove the comment if it does not need to be the first statement in the function, or move this line to be the first

Yotam7 (approved):

src/main/java/frc/robot/RobotManager.java
```diff
@@ -69,7 +72,9 @@ public void teleopInit() {
 
 	@Override
 	public void robotPeriodic() {
-		TimeUtils.updateCycleTime(); // Better to be first
+		roborioCycles++; // Better to be first
+		Logger.recordOutput("RoborioCycles", roborioCycles); // Better to be second
```
YHerm, line 76:
```suggestion
		Logger.recordOutput("RoborioCycles", roborioCycles); 
```

src/main/java/frc/utils/time/TimeUtils.java
```diff
@@ -34,6 +35,7 @@ public static void updateCycleTime() {
 	private static void logStatus() {
 		Logger.recordOutput(TimeConstants.LOG_PATH + "CycleTimeSeconds", getCurrentCycleTimeSeconds());
 		Logger.recordOutput(TimeConstants.LOG_PATH + "CurrentTimeSeconds", getCurrentTimeSeconds());
+		Logger.recordOutput(TimeConstants.LOG_PATH + "AverageCycleTimeSeconds", getCurrentTimeSeconds() / RobotManager.roborioCycles);
```
YHerm, line 38: save current time as var and use it twice

src/main/java/frc/robot/RobotManager.java
```diff
@@ -69,7 +72,9 @@ public void teleopInit() {
 
 	@Override
 	public void robotPeriodic() {
-		TimeUtils.updateCycleTime(); // Better to be first
+		roborioCycles++; // Better to be first
+		TimeUtils.updateCycleTime(); // Better to be second
+		Logger.recordOutput("RoborioCycles", roborioCycles);
```
YHerm, line 77: bro just rm comment, of course it will be after changing the roborio cycles

src/main/java/frc/utils/time/TimeUtils.java
```diff
@@ -14,6 +15,8 @@ public class TimeUtils {
 	private static double lastCycleTimeSeconds = 0;
 	private static double newCycleTimeSeconds = 0;
 
+	private static double currentTime;
```
YHerm, line 18: no as memvber

drorta: ?

src/main/java/frc/robot/RobotManager.java
drorta: do you mind if i just remove all the comments of "better to be _"

drorta: i dont think its neccessey

drorta: worst case you write a comment to move something if its there

YHerm: i prefer the comments to be there, i and everyone will not rmember to "worst case you write a comment to move something if its there"

drorta: so just a comment on the roborio++ to be first?

src/main/java/frc/utils/time/TimeUtils.java
YHerm: don't save it as member, save in the function to remove redundant calls for the same func

src/main/java/frc/robot/RobotManager.java
YHerm: and the time

src/main/java/frc/utils/time/TimeUtils.java
drorta: fine

src/main/java/frc/robot/RobotManager.java
drorta: both better to be first?

src/main/java/frc/robot/RobotManager.java
```diff
@@ -69,7 +72,9 @@ public void teleopInit() {
 
 	@Override
 	public void robotPeriodic() {
+		roborioCycles++; // Better to be first
 		TimeUtils.updateCycleTime(); // Better to be first
+		Logger.recordOutput("RoborioCycles", roborioCycles);
```
YHerm, line 77: put this above the time, in time say better to be second in comment

src/main/java/frc/utils/time/TimeUtils.java
```diff
@@ -32,8 +33,11 @@ public static void updateCycleTime() {
 	}
 
 	private static void logStatus() {
+		double currentTime = getCurrentTimeSeconds();
```
YHerm, line 36:
```suggestion
		double currentTimeSeconds = getCurrentTimeSeconds();
```

hillelv (approved):

YHerm (approved):
make sure its tested (robot or sim i dont really care), also update with core WOOW THAT WAS FAST

src/main/java/frc/robot/RobotManager.java
```diff
@@ -69,7 +72,9 @@ public void teleopInit() {
 
 	@Override
 	public void robotPeriodic() {
-		TimeUtils.updateCycleTime(); // Better to be first
+		roborioCycles++; // Better to be first
+		Logger.recordOutput("RoborioCycles", roborioCycles);
```
YoniKiriaty, line 76: Shouldn't this be in TimeUtils?

src/main/java/frc/robot/RobotManager.java
```diff
@@ -24,6 +25,8 @@
  */
 public class RobotManager extends LoggedRobot {
 
+	public static int roborioCycles;
```
YoniKiriaty, line 28: Consider making this not static

YoniKiriaty, line 28: You should probably initialize this value at 0, currently it will through a runtime error

YoniKiriaty (changes requested):

src/main/java/frc/robot/RobotManager.java
drorta: it doesnt have anything to do with time

src/main/java/frc/robot/RobotManager.java
drorta: i need to address it from timeutils so it has to be

drorta: it works, im pretty sure 0 its just the default value

src/main/java/frc/robot/RobotManager.java
```diff
@@ -69,7 +72,9 @@ public void teleopInit() {
 
 	@Override
 	public void robotPeriodic() {
-		TimeUtils.updateCycleTime(); // Better to be first
+		roborioCycles++; // Better to be first
+		Logger.recordOutput("RoborioCycles", roborioCycles);
+		TimeUtils.updateCycleTime(); // Better to be second
```
Yotam7, line 77: the comment states that this should be second, but it is not the second line of the function. Do one of the following:
1. move to be the second line
2. change comment to `//Better to be third`
3. clarify the meaning of `second` in the comment(something like `// Better to be second non logging statement` [this specific comment is rather confusing, so you should probably try to think of something clearer])

drorta: i dont think its neccessery to do any of these
i mean its kinda like common sense to log something after you change it and i dont think we should add any more comments
in my opinion its just fine like this

src/main/java/frc/robot/RobotManager.java
Yotam7: you are correct, `roborioCycles` will be initialized with `int`'s default value(`0`). However I do think you should be explicitly setting `roborioCycles` to `0`, for improved code clarity

YHerm: ^
 |

src/main/java/frc/robot/RobotManager.java
Yotam7: maybe move logging of `roborioCycles` to reside with all other loggings (after `CommandScheduler.getInstance().run();`

YHerm: i think its better to log right after change in the value

src/main/java/frc/robot/RobotManager.java
drorta: I don't think it's necessary I think it's pretty clear that way and that would just be like x=x no point in that

Yotam7 (dismissed):

Nitay4 (approved):

src/main/java/frc/robot/RobotManager.java
YoniKiriaty: It very much has something to do with time

src/main/java/frc/robot/RobotManager.java
YoniKiriaty: this ruins the design of `RobotManager` not being a singleton

src/main/java/frc/robot/RobotManager.java
YoniKiriaty: the function in `TimeUtils` is even called `updateCycleTime()`

src/main/java/frc/robot/RobotManager.java
YHerm: i dont think its ruining the design, why do u think so?

src/main/java/frc/robot/RobotManager.java
YHerm: i personally think its robot related (like very much only general robot related), and useful in time, and that robot manager is a great place for it

YHerm (dismissed):

src/main/java/frc/robot/RobotManager.java
drorta: @YoniKiriaty

src/main/java/frc/robot/RobotManager.java
drorta: @YoniKiriaty

drorta: I'm pretty sure that's the point of static

GilStein1: Yea, but if it didn't get to the part that it gives it a value, then exceptions will be thrown since you would be trying to access an int without a value

GilStein1: Usually, when we have static values they are final. It really is a problem if such thing happens in other classes. Usually there is no problem with that since the code is built in a way that it gets initialized before other classes get to use it but since this is robot template and the people of the future would like to use it, we cant make sure they will not change something in the code that would use the int before it initializes

src/main/java/frc/utils/time/TimeUtils.java
```diff
@@ -32,8 +33,11 @@ public static void updateCycleTime() {
 	}
 
 	private static void logStatus() {
+		double currentTimeSeconds = getCurrentTimeSeconds();
+
 		Logger.recordOutput(TimeConstants.LOG_PATH + "CycleTimeSeconds", getCurrentCycleTimeSeconds());
-		Logger.recordOutput(TimeConstants.LOG_PATH + "CurrentTimeSeconds", getCurrentTimeSeconds());
+		Logger.recordOutput(TimeConstants.LOG_PATH + "CurrentTimeSeconds", currentTimeSeconds);
+		Logger.recordOutput(TimeConstants.LOG_PATH + "AverageCycleTimeSeconds", currentTimeSeconds / RobotManager.roborioCycles);
```
YHerm, line 40: make this function get roborioCycles as param instead if using it secretly like this

src/main/java/frc/robot/RobotManager.java
YHerm, line 28: make it private and not static

src/main/java/frc/utils/time/TimeUtils.java
```diff
@@ -24,16 +24,19 @@ public class TimeUtils {
 		);
 	}
 
-	public static void updateCycleTime() {
+	public static void updateCycleTime(double roborioCycles) {
```
YoniKiriaty, line 27: imt

YoniKiriaty (changes requested):

src/main/java/frc/robot/hardware/ConnectedInput.java
```diff
@@ -4,7 +4,6 @@
 
 @AutoLog
 public class ConnectedInput {
-
```
Yotam7, line 7: You don't pass because of this removal

YHerm (approved):

hillelv (approved):

Yotam7 (approved):

YoniKiriaty (approved):

bossyshrimp68 (approved):

PR #441, opened 2024-11-05
Maple sim

YHerm:


src/main/java/frc/utils/joysticks/SmartJoystick.java
```diff
@@ -8,7 +8,7 @@
 
 public class SmartJoystick {
 
-	private static final double DEADZONE = 0.07;
```
Sh-Pe, line 11: does it have anything to do with the simulation?

Sh-Pe (commented):
does mapleSim work only with the latest beta versions of everything? Including Gralde?

Sh-Pe:
That's huge, thanks for implementing this.

src/main/java/frc/robot/poseestimation/posecalculation/PoseCalculator.java
```diff
@@ -37,13 +36,11 @@ public class PoseCalculator {
...
-			new SwerveModulePosition(),
-			new SwerveModulePosition(),
-			new SwerveModulePosition(),
-			new SwerveModulePosition()}
-	);
+	private SwerveModulePosition[] lastWheelPositions = new SwerveModulePosition[] {
+		new SwerveModulePosition(),
+		new SwerveModulePosition(),
+		new SwerveModulePosition(),
+		new SwerveModulePosition()};
```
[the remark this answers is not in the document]
YHerm: formatter doesnt approves this

src/main/java/frc/robot/subsystems/swerve/factories/modules/SimulationModuleGenerator.java
```diff
@@ -0,0 +1,31 @@
...
+
+import java.util.function.Supplier;
+
+public class SimulationModuleGenerator {
+
+	private static final DCMotor DRIVE_MOTOR = DCMotor.getFalcon500(1);
+	private static final DCMotor STEER_MOTOR = DCMotor.getFalcon500(1);
+	private static final double DRIVE_CURRENT_LIMIT_AMPS = 60;
+	private static final SwerveModuleSimulation.DRIVE_WHEEL_TYPE DRIVE_WHEEL_TYPE = SwerveModuleSimulation.DRIVE_WHEEL_TYPE.RUBBER;
+	private static final int GEAR_RATIO_LEVEL = 3; // L3
```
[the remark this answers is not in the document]
YHerm: L3 is a sds swerve module gearing (u can search on it) as the number (L1, L2, L3,...) gets higher u loose acceleration and gain velocity

src/main/java/frc/robot/subsystems/swerve/module/Module.java
```diff
@@ -199,37 +101,35 @@ public void setTargetOpenLoopVelocity(double targetVelocityMetersPerSecond) {
...
-		return driveInputs.velocityMetersPerSecond;
+	public Rotation2d getDriveLatestPosition() {
+		Rotation2d[] drivePositions = getDrivePositions();
+		return drivePositions[drivePositions.length - 1];
 	}
 
-	public Rotation2d getSteerPosition() {
-		return steerStuff.positionSignal().getLatestValue();
+	public double getDriveVelocityMetersPerSecond() {
+		return toDriveMeters(getDriveVelocitySeconds());
```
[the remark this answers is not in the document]
YHerm: what do u mean?

src/main/java/frc/robot/poseestimation/posecalculation/PoseCalculator.java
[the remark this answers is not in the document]
Sh-Pe: then formatter:off -- on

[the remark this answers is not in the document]
Sh-Pe: then whatever the formatter wants, till we'll change its behavior, if we will.

src/main/java/frc/robot/Robot.java
```diff
@@ -25,18 +35,31 @@
...
 	public Robot() {
+		GyroSimulation gyroSimulation = null;
+		if (ROBOT_TYPE.isSimulation()) {
+			gyroSimulation = SimulationGyroConstants.generateGyroSimulation();
+			Supplier<SwerveModuleSimulation> simulationModule = SimulationModuleGenerator.generate();
+			this.swerveDriveSimulation = SimulationSwerveGenerator
+				.generate(simulationModule, gyroSimulation, PoseEstimatorConstants.DEFAULT_POSE);
+			SimulatedArena.getInstance().addDriveTrainSimulation(swerveDriveSimulation);
+		} else {
+			swerveDriveSimulation = null;
```
Sh-Pe, line 55: Use optional instead, and pass it as an argument to the factory.

src/main/java/frc/robot/poseestimation/posecalculation/PoseCalculator.java
```diff
@@ -16,7 +16,6 @@
 import edu.wpi.first.math.geometry.Twist2d;
```
Sh-Pe, line 16: This is really old code of poseestimator. We're not using any of it by now. Since you'll probably merge this first, this is fine, but keep in mind that these changes will be deleted.

src/main/java/frc/robot/subsystems/swerve/module/hardware/HardwareModule.java
```diff
@@ -0,0 +1,163 @@
...
+
+public class HardwareModule extends Module {
+
+	@AutoLog
+	public static class CouplingInputs {
+
+		public Rotation2d uncoupledVelocityPerSecond = new Rotation2d();
+		public Rotation2d[] uncoupledPositions = new Rotation2d[0];
+
+	}
```
Sh-Pe, line 27: Same here, it's fine if we decide it would be a part of the module, but we need to be consistent -- and right now we're doing the opposite.

src/main/java/frc/robot/subsystems/swerve/module/Module.java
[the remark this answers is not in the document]
Sh-Pe: from what I've seen, you're using a different implementation (now with `uncoupleDriveAngle` too) so it may break the odometry. I would test it with a recent version of poseestimator.

src/main/java/frc/robot/subsystems/swerve/module/maple/MapleModule.java
```diff
@@ -0,0 +1,123 @@
...
+import org.ironmaple.simulation.drivesims.SwerveModuleSimulation;
+import org.littletonrobotics.junction.AutoLog;
+import org.littletonrobotics.junction.Logger;
+
+public class MapleModule extends Module {
+
+	@AutoLog
+	public static class ModuleIOInputs {
+
+		public double drivePositionRad = 0.0;
```
Sh-Pe, line 19: image

src/main/java/frc/robot/subsystems/swerve/module/maple/MapleModule.java
```diff
@@ -0,0 +1,123 @@
...
+		public double driveVelocityRadPerSec = 0.0;
```
Sh-Pe, line 20: image

src/main/java/frc/robot/subsystems/swerve/module/maple/MapleModule.java
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	@Override
+	public SysIdCalibrator.SysIdConfigInfo getDriveSysIdConfigInfo() {
+		return new SysIdCalibrator.SysIdConfigInfo(new SysIdRoutine.Config(), false);
+	}
+
+	@Override
+	public void updateInputs() {
+		moduleIOInputs.drivePositionRad = moduleSimulation.getDriveWheelFinalPositionRad();
```
Sh-Pe, line 59: I looked at their source code. Since the simulation uses 250Hz odometery for 50Hz roborio, I'd except to be able to get more than one value. `getCachedDriveEncoderUnGearedPositionsRad` should return a cache of the previous positions. It looks like the gear ratio isn't not applied on it since it caches the raw data from the encoder.

src/main/java/frc/robot/Robot.java
YHerm: Now I do it but without optional

src/main/java/frc/robot/subsystems/swerve/module/Module.java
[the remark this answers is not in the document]
Sh-Pe: I don't thinks so, if you're sure the data is exactly the same it's fine, but last time you've done major changes to the swerve that changed the odometry, it took a while to debug it. So I think it's better to test each change to the swerve by itself, to make sure it doesn't happen again.

build.gradle
```diff
@@ -3,7 +3,7 @@ import groovy.json.JsonSlurper
 
 plugins {
 	id "java"
-	id "edu.wpi.first.GradleRIO" version "2024.3.2"
+	id "edu.wpi.first.GradleRIO" version "2025.1.1-beta-1"
```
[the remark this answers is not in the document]
YHerm: this is the version that works with maple-sim, I know this shouldn't be part of this pr. We didnt update versions in the main code now

src/main/java/frc/robot/subsystems/swerve/module/Module.java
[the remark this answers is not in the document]
YHerm: i just added a step in function in the middle instead of two steps in same func. if u see something specific here that maybe i didnt notice let me know

src/main/java/frc/robot/subsystems/swerve/module/maple/MapleModule.java
```diff
@@ -0,0 +1,123 @@
...
+
+	@Override
+	public void setTargetSteerPosition(Rotation2d targetSteerPosition) {
+		setSteerVoltage(steerPIDControllerRadians.calculate(getSteerLatestPosition().getRadians(), targetSteerPosition.getRadians()));
+	}
+
+	@Override
+	public void setTargetClosedLoopVelocity(double targetVelocityMetersPerSecond) {
+		Rotation2d targetVelocityRadiansPerSecond = fromDriveMeters(targetVelocityMetersPerSecond);
+		double voltage = driveFeedForwardRadiansPerSecond.calculate(targetVelocityRadiansPerSecond.getRadians())
```
Sh-Pe, line 98: looks like it's deprecated... I get the following warning:
> ```Note: Some input files use or override a deprecated API.```

src/main/java/frc/robot/subsystems/swerve/module/hardware/HardwareModule.java
```diff
@@ -0,0 +1,163 @@
...
+
+	private final IAngleEncoder encoder;
+	private final EncoderStuff encoderStuff;
```
tnrz135, line 30: Try to avoid having to call something: "*Stuff"

src/main/java/frc/robot/Robot.java
tnrz135: There is no need for this here, the value can simply stay null as it is from the start. You simply don't need the else section

tnrz135 (changes requested):

src/main/java/frc/robot/Robot.java
YHerm: Because of the final the else is needed

src/main/java/frc/robot/subsystems/swerve/module/maple/MapleModule.java
YHerm: you're right its deprecated but the fix is not for this pr

src/main/java/frc/robot/subsystems/swerve/module/hardware/HardwareModule.java
Sh-Pe: It's a convention in our team, that is being used in core quite a lot. There were discussions around this name, from what I know.

src/main/java/frc/robot/subsystems/swerve/factories/modules/SimulationModuleGenerator.java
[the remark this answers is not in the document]
Sh-Pe: So maybe change it to `SDS_GEAR_RATIO_LEVEL` since it isn’t a thing outside to the sds ecosystem.

Sh-Pe (changes requested):
You're using an older version of maple sim. The latest version is 0.1.6 (json [here](https://shenzhen-robotics-alliance.github.io/maple-sim/vendordep/maple-sim.json)). Note that they've broken some stuff.

PR #456, opened 2024-11-15
SparkMaxRequest gets rps -> core

drorta:


YHerm (approved):
Good job, Make sure to test it!

src/main/java/frc/robot/hardware/rev/request/SparkMaxRequest.java
```diff
@@ -42,8 +44,12 @@ public T getSetPoint() {
 		return setPoint;
 	}
 
-	public Double getSetPointAsDouble() {
-		return setPointToDoubleConverter.apply(setPoint);
+	public Double getSetPointForSparkMAX() {
```
hillelv, line 47: I don't like the name ForSparkMAX because the entire class is for spark max, and the name doesn't tell me what the function does to the set point. I don't think I would understand what this function is without looking inside of it

drorta: I think it's pretty clear, the forsparkmax returns the setpoint for the SparkMax, the setpoint that is entered to the motor

YHerm: U can do getSparkMaxCompatibleSetPoint or something like it if u want

drorta: Funny but yeah sounds more right

hillelv (approved):

YHerm (approved):
make sure its tested!

Sh-Pe (approved):

bossyshrimp68 (approved):

PR #472, opened 2024-11-23
Choreo Vendor Update -> core

YHerm:


hillelv:
you can delete the vendor to remove the error for now

PR #476, opened 2024-11-26
Time Related Data (cr fix) -> core

YHerm:


drorta (approved):
my stylistic choice would be to put the method after the periodic, your choice

drorta (dismissed):
make sure its tested

drorta (approved):

PR #513, opened 2024-12-15
update

YHerm:


Sh-Pe (changes requested):
You just deleted @emilyyyylime'sformatter and run the default one

I get that the previous gradle had compile issues in the latest versions but I don’t think it's replayed to the formatter any anyway we should scrap it.

Sh-Pe:
Spaces over tabs has gone too far

hillelv:
This is just the first commit that copies the build file from maple, all the stuff from our build will return before merge

Sh-Pe (changes requested):

PR #536, opened 2024-12-24
RobotHeadingEstimator -> poseestimator-master

GilStein1:
Limelight has 2 important pose estimating features:

*megatag1
*megatag2

megatag1 is the old one, which is not much accurate and has a lot of noise.
megatag2 is the new one, which is much, much better than the old one, but it requires that we send the robots heading to the limelight.

Since we want to use the megatag2, we need a system for finding out the offset between the gyro's yaw and the actual heading of the robot, and then to send it to the limelight.

The purpose of this class is to estimate the robot heading by combining the gyro's output and the megatag1 heading estimation using the Kalman filter algorithm.

This class would probably be used by the MultiApriltagVisionSources.


src/main/java/frc/robot/poseestimator/PoseEstimatorMath.java
```diff
@@ -0,0 +1,38 @@
...
+		double visionStandardDeviation
+	) {
+		double combinedStandardDeviation = getKalmanRatio(gyroStandardDeviation, visionStandardDeviation);
+		Rotation2d estimatedHeadingAtSampleTime = currentEstimatedHeading.plus(sampleDifferenceFromHeading);
+		Rotation2d differenceFromVisionAndEstimatedHeading = getShortestAngleDifference(visionEstimatedHeading, estimatedHeadingAtSampleTime);
+		Rotation2d scaledDifferenceToAddToEstimatedHeading = differenceFromVisionAndEstimatedHeading.times(combinedStandardDeviation);
+		return currentEstimatedHeading.plus(scaledDifferenceToAddToEstimatedHeading);
+	}
+
+	public static Rotation2d getShortestAngleDifference(Rotation2d angle1, Rotation2d angle2) {
```
Sh-Pe, line 25:
```suggestion
	public static Rotation2d getAngleDistance(Rotation2d angle1, Rotation2d angle2) {
```
distance = shortest angle difference, in the context of a circle.

src/main/java/frc/robot/poseestimator/helpers/RobotHeadingEstimator.java
```diff
@@ -0,0 +1,48 @@
...
+			)
+		);
+	}
+
+	public void updateGyroAngle(Rotation2d heading, double timestamp) {
+		gyroAngleInterpolator.addSample(timestamp, heading);
+		estimatedHeading = estimatedHeading.plus(PoseEstimatorMath.getShortestAngleDifference(heading, lastGyroAngle));
+		lastGyroAngle = heading;
+	}
+
```
Sh-Pe, line 47: add a reset method. At the beginning you know that the gyro angle is not true, and you can initialize the offset using the PP.

src/main/java/frc/robot/poseestimator/helpers/RobotHeadingEstimator.java
```diff
@@ -0,0 +1,48 @@
...
+	public Rotation2d getEstimatedHeading() {
+		return estimatedHeading;
+	}
+
+	public void updateVisionHeading(Rotation2d heading, double timestamp) {
+		Optional<Rotation2d> gyroAtTimestamp = gyroAngleInterpolator.getSample(timestamp);
+		double visionHeadingStandardDeviation = PoseEstimatorMath
+			.calculateStandardDeviation(heading.getRadians(), estimatedHeading.getRadians());
```
Sh-Pe, line 30: why is this not calculated in the function call, but the difference does? Either everything is inline, or everything is separated.

src/main/java/frc/robot/poseestimator/PoseEstimatorMath.java
```diff
@@ -0,0 +1,38 @@
...
+
+import edu.wpi.first.math.geometry.*;
+
+public class PoseEstimatorMath {
+
+	public static double getKalmanRatio(double odometryStandardDeviation, double visionStandardDeviation) {
+		return odometryStandardDeviation / (odometryStandardDeviation + visionStandardDeviation);
+	}
+
+	public static Rotation2d combineVisionHeadingToGyro(
```
Sh-Pe, line 11:
```suggestion
	public static Rotation2d combineOldVisionHeadingToGyro(
```

at least older than the gyro measurement

src/main/java/frc/robot/poseestimator/PoseEstimatorMath.java
```diff
@@ -0,0 +1,38 @@
...
+		Rotation2d visionEstimatedHeading,
+		Rotation2d sampleDifferenceFromHeading,
```
Sh-Pe, line 13:
```suggestion
		Rotation2d changeInAngleSinceVisionObservationWasTaken,
```

I would move the calculation of it to the function body itself, since it's anyway computed using what anyways given to the function, and it matches its responsibility

src/main/java/frc/robot/poseestimator/PoseEstimatorMath.java
```diff
@@ -0,0 +1,38 @@
...
+		Rotation2d currentEstimatedHeading,
+		double gyroStandardDeviation,
+		double visionStandardDeviation
+	) {
+		double combinedStandardDeviation = getKalmanRatio(gyroStandardDeviation, visionStandardDeviation);
+		Rotation2d estimatedHeadingAtSampleTime = currentEstimatedHeading.plus(sampleDifferenceFromHeading);
```
Sh-Pe, line 19: So you've done this calculation: 
```estimatedHeadin - |estimatedHeading - gyroSample|```

Which is just the gyrosmaple. 
I think you want to add the difference between the samples and not just using the gyro estimation at timestamp.

File: src/main/java/frc/robot/poseestimator/PoseEstimatorMath.java [hunk 7]
```diff
@@ -0,0 +1,38 @@
...
+		Rotation2d differenceFromVisionAndEstimatedHeading = getShortestAngleDifference(visionEstimatedHeading, estimatedHeadingAtSampleTime);
+		Rotation2d scaledDifferenceToAddToEstimatedHeading = differenceFromVisionAndEstimatedHeading.times(combinedStandardDeviation);
```
Sh-Pe, line 21: This is _not_ how you use a ratio, nor stdDev. The point of a ratio is that too parts adds up to one, proportional to the ratio. This completely breaks the idea of units. 

If you trust "more", WOLOG, the vision, then you want to give more weight to the vision (e.g. using a weighted mean of the gyro difference and the vision difference from current heading), rather than moving the robot/heading more "far away" from the current heading.

src/main/java/frc/robot/poseestimator/PoseEstimatorMath.java
```diff
@@ -0,0 +1,38 @@
...
+	}
+
+	public static Rotation2d combineVisionHeadingToGyro(
+		Rotation2d visionEstimatedHeading,
+		Rotation2d sampleDifferenceFromHeading,
+		Rotation2d currentEstimatedHeading,
+		double gyroStandardDeviation,
+		double visionStandardDeviation
+	) {
+		double combinedStandardDeviation = getKalmanRatio(gyroStandardDeviation, visionStandardDeviation);
```
Sh-Pe, line 18: stdDev or ratio?

Sh-Pe (commented):
Happy Hannukah

src/main/java/frc/robot/poseestimator/helpers/RobotHeadingEstimator.java
```diff
@@ -0,0 +1,48 @@
...
+
+import edu.wpi.first.math.geometry.Rotation2d;
+import edu.wpi.first.math.interpolation.TimeInterpolatableBuffer;
+import frc.robot.poseestimator.PoseEstimatorMath;
+
+import java.util.Optional;
+
+public class RobotHeadingEstimator {
+
+	private final TimeInterpolatableBuffer<Rotation2d> gyroAngleInterpolator;
```
Sh-Pe, line 11: perhaps you should clarify in the name of the the interpolator that it's unoffseted gyro angle and can't be rawly used for anything outside of calculating offsets.

NoamRosenberg08 (approved):

src/main/java/frc/robot/poseestimator/PoseEstimatorMath.java
```diff
@@ -0,0 +1,38 @@
...
+
+public class PoseEstimatorMath {
+
+	public static double getKalmanRatio(double odometryStandardDeviation, double visionStandardDeviation) {
+		return odometryStandardDeviation / (odometryStandardDeviation + visionStandardDeviation);
+	}
+
+	public static Rotation2d combineVisionHeadingToGyro(
+		Rotation2d visionEstimatedHeading,
+		Rotation2d changeInAngleSinceVisionObservationWasTaken,
```
asaf8888, line 13: This claims this is the change in angle since the vision observation however you seem to be using it's inverse (you are passing into it angleatobservation.minus(lastangle)) the math checks out but the name would make more sense if you changed the function to actually take the change

src/main/java/frc/robot/poseestimator/helpers/RobotHeadingEstimator.java
```diff
@@ -0,0 +1,52 @@
...
+				heading,
+				PoseEstimatorMath.getAngleDistance(gyroSampleAtTimestamp, lastGyroAngle),
+				estimatedHeading,
+				gyroStandardDeviation,
+				visionStandardDeviation
+			)
+		);
+	}
+
+	public void updateGyroAngle(Rotation2d heading, double timestamp) {
```
asaf8888, line 46: To use this you would have to know to first call update gyro and then update by vision shouldn't there be one function that does both in the right order?

src/main/java/frc/robot/poseestimator/PoseEstimatorMath.java
```diff
@@ -0,0 +1,38 @@
...
+	public static Rotation2d getAngleDistance(Rotation2d angle1, Rotation2d angle2) {
+		Rotation2d difference = angle1.minus(angle2);
+		if (difference.getRadians() > Math.PI) {
+			return Rotation2d.fromRadians(2 * Math.PI - difference.getRadians());
+		}
+		return difference;
+	}
+
+	public static double calculateStandardDeviation(double estimatedValue, double currentValue) {
+		double mean = (estimatedValue + currentValue) / 2;
```
DANNA2828, line 34: Average

src/main/java/frc/robot/poseestimator/helpers/RobotHeadingEstimator.java
GilStein1: The order is not really important

Sh-Pe: I think he meant to ask for another update func that combines both

src/main/java/frc/robot/poseestimator/PoseEstimatorMath.java
Sh-Pe: So `changeInGyroHeadingFromWhenVisionObservationWasTakenToNow`? 
at this point it's better to pass the gyrosampleattiemstamp and the gyrosample both into that func, which as I (think I) stated somewhere above, is better in my opinion because it is a math thing (-> poseestimatormath)

asaf8888: I am actually suggesting to leave the name the same but flip the use as I think it is more intuitive anyways but perhaps you are right

GilStein1: I already fixed it, kept the name but reversed the usage👍

src/main/java/frc/robot/poseestimator/helpers/RobotHeadingEstimator.java
```diff
@@ -0,0 +1,81 @@
...
+		return estimatedHeading;
+	}
+
+	public void updateVisionHeading(HeadingData visionHeadingData, double visionStandardDeviation) {
+		angleBuffer.addAngle(visionHeadingData.heading());
+
+		Optional<Rotation2d> slidingWindowAngleAccumulatorAverage = angleBuffer.average();
+		if (
+			slidingWindowAngleAccumulatorAverage.isPresent()
+				&& PoseEstimatorMath.getAngleDistance(slidingWindowAngleAccumulatorAverage.get(), visionHeadingData.heading()).getRadians()
```
GilStein1, line 44: I would put an absolute value in there because the Rotation2d put the values between 180 and -180

src/main/java/frc/robot/poseestimator/helpers/RingBufferIterator.java
```diff
@@ -0,0 +1,39 @@
...
+import java.util.Iterator;
+
+public class RingBufferIterator<T> implements Iterator<T> {
+
+	private final RingBuffer<T> ringBuffer;
+	private final int endIndex;
+	private int currentIndex;
+	private boolean used;
+
+	RingBufferIterator(RingBuffer<T> ringBuffer) {
```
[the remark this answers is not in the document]
Yotam7: default is more restrictive than `protected`, why give more access?

src/main/java/frc/robot/poseestimator/helpers/RingBufferIterator.java
```diff
@@ -0,0 +1,39 @@
...
+
+	@Override
+	public boolean hasNext() {
+		return ringBuffer.existsAtIndex(currentIndex) && !((this.used) && currentIndex == endIndex);
+	}
+
+	@Override
+	public T next() {
+		if (!hasNext()) {
+			return null;
```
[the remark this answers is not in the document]
Yotam7: Since this function is implementing an `interface`, you cannot throw an exception (the interface didn't specify it so it's not allowed)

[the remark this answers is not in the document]
Sh-Pe: it defiantly *does* work. I'm pretty sure it's because it's a runtimeexception
20250119_19h39m39s_grim

src/main/java/frc/robot/constants/RobotHeadingEstimatorConstants.java
```diff
@@ -0,0 +1,17 @@
+package frc.robot.constants;
+
+import edu.wpi.first.math.geometry.Rotation2d;
+
+public class RobotHeadingEstimatorConstants {
+
+	public static final double POSE_BUFFER_SIZE_SECONDS = 2.0;
+
+	public static final double DEFAULT_GYRO_STANDARD_DEVIATION = 0.0001;
```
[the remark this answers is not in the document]
GilStein1: ?
what?

src/main/java/frc/constants/RobotHeadingEstimatorConstants.java
```diff
@@ -0,0 +1,18 @@
+package frc.constants;
+
+
+public class RobotHeadingEstimatorConstants {
+
+	public static final double POSE_BUFFER_SIZE_SECONDS = 2.0;
+
+	public static final double AMOUNT_OF_SOURCE_TYPES = 2.0;
```
Stormageddon37, line 8: How is an amount a double?

src/main/java/frc/robot/poseestimator/PoseEstimatorMath.java
```diff
@@ -0,0 +1,38 @@
...
+
+	public static Rotation2d combineVisionHeadingToGyro(
+		Rotation2d visionEstimatedHeading,
+		Rotation2d gyroAngle,
+		Rotation2d lastGyroAngle,
+		Rotation2d currentEstimatedHeading,
+		double gyroStandardDeviation,
+		double visionStandardDeviation
+	) {
+		Rotation2d changeInAngleSinceVisionObservationWasTaken = PoseEstimatorMath.getAngleDistance(gyroAngle, lastGyroAngle);
```
Stormageddon37, line 22:
```suggestion
		Rotation2d changeInAngleSinceVisionObservation = PoseEstimatorMath.getAngleDistance(gyroAngle, lastGyroAngle);
```

src/main/java/frc/robot/poseestimator/PoseEstimatorMath.java
```diff
@@ -0,0 +1,38 @@
...
+		Rotation2d scaledDifferenceToAddToEstimatedHeading = differenceFromVisionAndEstimatedHeading.times(visionAndGyroRatio);
+		return currentEstimatedHeading.plus(scaledDifferenceToAddToEstimatedHeading);
+	}
+
+	public static Rotation2d getAngleDistance(Rotation2d angle1, Rotation2d angle2) {
+		Rotation2d difference = angle1.minus(angle2);
+		if (difference.getRadians() > Math.PI) {
+			return Rotation2d.fromRadians(MathConstants.FULL_CIRCLE.getRadians() - difference.getRadians());
+		}
+		return difference;
```
Stormageddon37, line 35: Ternary statement?

src/main/java/frc/utils/AngleUtils.java
```diff
@@ -0,0 +1,16 @@
+package frc.utils;
+
+import edu.wpi.first.math.geometry.Rotation2d;
+import frc.constants.MathConstants;
+
+public class AngleUtils {
+
+	public static Rotation2d wrappingAbs(Rotation2d angle) {
```
Stormageddon37, line 8: Abs? Like stomach muscles?

Stormageddon37 (changes requested):

hillelv (commented):
4 files out of 9

src/main/java/frc/utils/buffers/RingBuffer/RingBuffer.java
```diff
@@ -0,0 +1,83 @@
...
+			index += size();
+		}
+		return index;
+	}
+
+	protected int getInsertions() {
+		return insertions;
+	}
+
+	protected boolean existsAtIndex(int index) {
```
hillelv, line 70:
```suggestion
	protected boolean isExist(int index) {
```

src/main/java/frc/utils/buffers/RingBuffer/RingBuffer.java
```diff
@@ -0,0 +1,83 @@
...
+		return buffer.get(index).isPresent();
+	}
+
+	protected Optional<T> getAtIndex(int index) {
```
hillelv, line 74:
```suggestion
	protected Optional<T> get(int index) {
```

src/main/java/frc/robot/poseestimator/helpers/RobotHeadingEstimator.java
```diff
@@ -0,0 +1,110 @@
...
+	public Rotation2d getEstimatedHeading() {
+		return estimatedHeading;
+	}
+
+	public void updateVisionIfNotCalibrated(
+		TimedValue<Rotation2d> visionHeadingData,
+		double visionStandardDeviation,
+		double maximumStandardDeviationTolerance
+	) {
+		double visionNoiseStandardDeviation = PoseMath.calculateStandardDeviations(
```
Sh-Pe, line 50: Suggestion: since stdDev in WPILib's terminology is just an arbitrary function that describes how much we trust an observation, and not a well defined term, consider having that function as a constant that could be changed instead of a method.

src/main/java/frc/robot/poseestimator/PoseEstimatorMath.java
```diff
@@ -0,0 +1,30 @@
...
+
+import edu.wpi.first.math.geometry.Rotation2d;
+import frc.constants.RobotHeadingEstimatorConstants;
+import frc.utils.math.AngleMath;
+
+public class PoseEstimatorMath {
+
+	public static double getKalmanRatio(double odometryStandardDeviation, double visionStandardDeviation) {
+		double ratio = odometryStandardDeviation / (odometryStandardDeviation + visionStandardDeviation);
+		return Double.isInfinite(ratio) ? 1.0 / RobotHeadingEstimatorConstants.AMOUNT_OF_SOURCE_TYPES : ratio;
```
asaf8888, line 11: What is this calculation?
It is completely unintuitive

asaf8888: Or I mean to say it is not obvious that's how it will behave and also don't fix it by changing the name it at the very least should happen in another function and more likely not at all

asaf8888: Also why specifically the number of sources

src/main/java/frc/robot/poseestimator/PoseEstimatorMath.java
```diff
@@ -0,0 +1,30 @@
...
+
+	public static Rotation2d combineVisionHeadingAndGyro(
+		Rotation2d visionEstimatedHeading,
+		Rotation2d gyroAngle,
+		Rotation2d lastGyroAngle,
+		Rotation2d currentEstimatedHeading,
+		double gyroStandardDeviation,
+		double visionStandardDeviation
+	) {
+		Rotation2d changeInAngleSinceVisionDataWasObserved = AngleMath.getAngleDifference(gyroAngle, lastGyroAngle);
```
asaf8888, line 22: This assumes the last gyro angle was taken at the same time as the vision observation it is either incorrect or misnamed

src/main/java/frc/utils/math/PoseMath.java
```diff
@@ -13,4 +15,19 @@ public static Translation2d getRelativeTranslation(Pose2d relativeTo, Translatio
 		return getRelativeTranslation(relativeTo.getTranslation(), toRelative).rotateBy(relativeTo.getRotation().unaryMinus());
 	}
 
+	public static <T extends Iterable<E>, E> double calculateStandardDeviations(T values, Function<E, Double> getValue) {
```
asaf8888, line 18: Where is this used?

GilStein1: In RobotHeadingEstimator. There is a method for updating the vision if the heading is not calibrated yet. It takes the difference between the estimatedHeading and the gyro heading and calculates thw stdDev of those differences over the couple of last observations and if the stdDev is small enough it stops updating the vision

asaf8888 (approved):

NoamRosenberg08 (approved):

PR #537, opened 2024-12-24
Package Order -> master

YHerm:


src/main/java/frc/joysticks/Axis.java
```diff
@@ -1,4 +1,4 @@
-package frc.utils.joysticks;
+package frc.joysticks;
```
hillelv, line 1: Why is joysticks outside of utils?

src/main/java/frc/JoysticksBindings.java
```diff
@@ -1,7 +1,8 @@
-package frc.robot;
+package frc;
```
hillelv, line 1: Why is bindings outside of joysticks?

src/main/java/frc/RobotManager.java
```diff
@@ -2,15 +2,13 @@
 // Open Source Software; you can modify and/or share it under the terms of
 // the WPILib BSD license file in the root directory of this project.
 
-package frc.robot;
+package frc;
```
hillelv, line 5: Why is robot manager outside of robot?

src/main/java/frc/robot/RobotConstants.java
hillelv, line 1: What other constants can be here?

tnrz135 (approved):

YoniKiriaty (approved):

drorta (approved):

PR #547, opened 2024-12-28
Levi-Code-Review-PR -> Levi-Robot-CR

Yehonatan-Levi:


src/main/java/frc/robot/hardware/phoenix6/Phoenix6Device.java
```diff
@@ -0,0 +1,94 @@
...
+	private void logSignals(InputSignal<?>... signals) {
+		for (InputSignal<?> signal : signals) {
+			Logger.processInputs(logPath, signal);
+		}
+	}
+
+	@Override
+	public void updateInputs(InputSignal<?>... inputSignals) {
+		InputSignal<?>[] validSignals = getValidSignals(inputSignals);
+		refreshSignals(validSignals);
```
Yehonatan-Levi, line 86: why are you not saving the return value? you can add an alert or something

src/main/java/frc/utils/Conversions.java
```diff
@@ -0,0 +1,187 @@
...
+
+public class Conversions {
+
+	public static final double MAG_TICKS = 4096;
+	public static final double HUNDRED_MILLISECONDS_PER_SECONDS = 10;
+	public static final double SECONDS_PER_MINUTE = 60;
+
+
+	/**
+	 * Converts ticks from a Mag Encoder to angle.
```
Yehonatan-Levi, line 15: why here there is documentation

src/main/java/frc/utils/Conversions.java
```diff
@@ -0,0 +1,187 @@
...
+		return perSecondToPerMinute(per100MillisecondsToPerSecond(frequency));
+	}
+
+	/**
+	 * Converts a frequency from per second to per minute.
+	 *
+	 * @param frequency the frequency per second
+	 * @return the frequency per minute
+	 */
+	public static double perSecondToPerMinute(double frequency) {
```
Yehonatan-Levi, line 84: why perSecondPerMinute isnt next to its counter part?

src/main/java/frc/utils/Conversions.java
Sh-Pe: Because it's a lib

src/main/java/frc/utils/time/TimeUtils.java
```diff
@@ -0,0 +1,48 @@
...
+import edu.wpi.first.hal.HALUtil;
+import frc.utils.Conversions;
+import frc.utils.alerts.Alert;
+import frc.utils.alerts.AlertManager;
+import frc.utils.alerts.PeriodicAlert;
+import org.littletonrobotics.junction.Logger;
+
+public class TimeUtils {
+
+	public static final double DEFAULT_CYCLE_TIME_SECONDS = 0.02;
```
Yehonatan-Levi, line 12: why not in TimeConstants?

YHerm: Constants that are useful out of this class should be in the open class.
If I want something about the time I always will go to time utils and no need to think time utils or time constants.
Constants are just for inner use without being exposed outside

PR #508, opened 2024-12-12
GBPoseEstimator ⇶ (in the future) master

Sh-Pe:
poseestimator

⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⠿⠛⠋⠉⠉⠁⠀⠀⠀⠀⠉⠉⠙⠛⠻⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿
⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⠟⠋⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢻⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿
⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⠏⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿
⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢸⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿
⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠘⠳⠄⢻⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿
⣿⣿⣿⣿⣿⣿⣿⣿⠿⠏⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠉⠙⠻⣿⣿⣿⣿⣿⣿⣿
[...]

PR #558, opened 2025-01-01
Swerve -> master

YHerm:


.git-blame-ignore-revs
drorta, line 1: check again about what to do with this

src/main/java/frc/JoysticksBindings.java
```diff
@@ -1,8 +1,8 @@
 package frc;
 
-import frc.robot.Robot;
 import frc.joysticks.JoystickPorts;
 import frc.joysticks.SmartJoystick;
+import frc.robot.Robot;
```
drorta, line 5: is this spotless? if not, i think should be in another pr its just funny here

src/main/java/frc/constants/GlobalConstants.java
```diff
@@ -5,5 +5,6 @@
 public class GlobalConstants {
 
 	public static final DriverStation.Alliance SIMULATION_ALLIANCE = DriverStation.Alliance.Blue;
+	public static final String SUBSYSTEM_LOG_PREFIX = "Subsystems/";
```
drorta, line 8: also maybe another pr

src/main/java/frc/robot/subsystems/swerve/Swerve.java
```diff
@@ -0,0 +1,297 @@
...
+	public SwerveStateHelper getStateHelper() {
+		return stateHelper;
+	}
+
+
+	public void configPathPlanner(Supplier<Pose2d> currentPoseSupplier, Consumer<Pose2d> resetPoseConsumer, RobotConfig robotConfig) {
+		try {
+			robotConfig = RobotConfig.fromGUISettings();
+		} catch (Exception exception) {
+			DriverStation.reportError(exception.getMessage(), exception.getStackTrace());
```
NoamRosenberg08, line 94: Maybe alerts/add a log here?

src/main/java/frc/JoysticksBindings.java
YHerm: u want to do it right?
(yes pack order changed so spotless)

NoamRosenberg08 (changes requested):

.git-blame-ignore-revs
YHerm: @emilyyyylime

drorta (dismissed):
nice job!!!

ItayKauff:
looks great

src/main/java/frc/robot/subsystems/swerve/Swerve.java
Sh-Pe: I agree. I think we can wrap the new Wpilib's alerts Manager together with ours, since their alerts (from what I've seen) don’t have logging. (By wrapping I mean creating a function that does both).

Sh-Pe:
Update yourself with `swerve-math`

src/main/java/frc/robot/subsystems/swerve/states/SwerveStateHelper.java
```diff
@@ -0,0 +1,75 @@
...
+			case MIDDLE_OF_CHASSIS -> new Translation2d();
+			case FRONT_LEFT_MODULE -> swerve.getModules().getModule(ModuleUtils.ModulePosition.FRONT_LEFT).getPositionFromCenterMeters();
+			case FRONT_RIGHT_MODULE -> swerve.getModules().getModule(ModuleUtils.ModulePosition.FRONT_RIGHT).getPositionFromCenterMeters();
+			case BACK_LEFT_MODULE -> swerve.getModules().getModule(ModuleUtils.ModulePosition.BACK_LEFT).getPositionFromCenterMeters();
+			case BACK_RIGHT_MODULE -> swerve.getModules().getModule(ModuleUtils.ModulePosition.BACK_RIGHT).getPositionFromCenterMeters();
+		};
+	}
+
+	public RotateAxis getFarRotateAxis(boolean isLeft) {
+		Rotation2d currentAllianceHeading = swerve.getAllianceRelativeHeading();
```
Sh-Pe, line 50:
```suggestion
		Rotation2d currentHeadingRelativeToAlliance = swerve.getAllianceRelativeHeading();
```
It has nothing to do with the heading of the alliance itself (whatever that means)

src/main/java/frc/robot/subsystems/swerve/states/SwerveStateHelper.java
```diff
@@ -0,0 +1,75 @@
...
+		return switch (rotationAxisState) {
+			case MIDDLE_OF_CHASSIS -> new Translation2d();
+			case FRONT_LEFT_MODULE -> swerve.getModules().getModule(ModuleUtils.ModulePosition.FRONT_LEFT).getPositionFromCenterMeters();
+			case FRONT_RIGHT_MODULE -> swerve.getModules().getModule(ModuleUtils.ModulePosition.FRONT_RIGHT).getPositionFromCenterMeters();
+			case BACK_LEFT_MODULE -> swerve.getModules().getModule(ModuleUtils.ModulePosition.BACK_LEFT).getPositionFromCenterMeters();
+			case BACK_RIGHT_MODULE -> swerve.getModules().getModule(ModuleUtils.ModulePosition.BACK_RIGHT).getPositionFromCenterMeters();
+		};
+	}
+
+	public RotateAxis getFarRotateAxis(boolean isLeft) {
```
Sh-Pe, line 49: I don’t get what this function is supposed to do. Maybe change the name.

src/main/java/frc/robot/subsystems/swerve/states/SwerveStateHelper.java
```diff
@@ -0,0 +1,75 @@
...
+
+import java.util.Optional;
+import java.util.function.Supplier;
+
+public class SwerveStateHelper {
+
+	private final Swerve swerve;
+	private final SwerveConstants swerveConstants;
+	private final Supplier<Optional<Pose2d>> robotPoseSupplier;
+	private final Supplier<Optional<Translation2d>> objectTranslationSupplier;
```
Sh-Pe, line 20: Never used

src/main/java/frc/robot/subsystems/swerve/Swerve.java
```diff
@@ -0,0 +1,297 @@
...
+		return Math.min(gyroSignals.yawSignal().asArray().length, modules.getNumberOfOdometrySamples());
+	}
+
+	public OdometryObservation[] getAllOdometryObservations() {
+		OdometryObservation[] odometryObservations = new OdometryObservation[getNumberOfOdometrySamples()];
+
+		for (int i = 0; i < odometryObservations.length; i++) {
+			odometryObservations[i] = new OdometryObservation(
+				modules.getWheelPositions(i),
+				gyro instanceof EmptyGyro ? null : gyroSignals.yawSignal().asArray()[i],
```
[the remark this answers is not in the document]
YHerm: Pose don’t get gyro it’s gets angle,
In simulation I calc angle with the wheels

src/main/java/frc/robot/subsystems/swerve/Swerve.java
```diff
@@ -0,0 +1,297 @@
...
+
+	private ChassisSpeeds getDriveModeRelativeSpeeds(ChassisSpeeds speeds, SwerveState swerveState) {
+		if (swerveState.getDriveMode() == DriveRelative.ROBOT_RELATIVE) {
+			return speeds;
+		}
+		return SwerveMath.fieldToRobotRelativeSpeeds(speeds, getAllianceRelativeHeading());
+	}
+
+
+	protected void moveToPoseByPID(Pose2d currentPose, Pose2d targetPose) {
```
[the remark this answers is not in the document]
YHerm: And not by path planner

src/main/java/frc/robot/subsystems/swerve/Swerve.java
```diff
@@ -0,0 +1,297 @@
...
+
+
+	public boolean isAtHeading(Rotation2d targetHeading, Rotation2d tolerance, Rotation2d velocityDeadbandAnglesPerSecond) {
+		double headingDeltaDegrees = Math.abs(targetHeading.minus(headingSupplier.get()).getDegrees());
+		boolean isAtHeading = headingDeltaDegrees < tolerance.getDegrees();
+
+		double rotationVelocityRadiansPerSecond = getRobotRelativeVelocity().omegaRadiansPerSecond;
+		boolean isStopping = Math.abs(rotationVelocityRadiansPerSecond) < velocityDeadbandAnglesPerSecond.getRadians();
+
+		return isAtHeading && isStopping;
```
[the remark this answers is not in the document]
YHerm: The angle velocity need to stop.
Else u can go throughout the angle and oscillate

src/main/java/frc/robot/subsystems/swerve/states/SwerveStateHelper.java
YHerm: Gives the X Y relative to mid of chassis of the module that far from driver

src/main/java/frc/robot/subsystems/swerve/Swerve.java
```diff
@@ -0,0 +1,297 @@
...
+
+	public SwerveStateHelper getStateHelper() {
+		return stateHelper;
+	}
+
+
+	public void configPathPlanner(Supplier<Pose2d> currentPoseSupplier, Consumer<Pose2d> resetPoseConsumer, RobotConfig robotConfig) {
+		try {
+			robotConfig = RobotConfig.fromGUISettings();
+		} catch (Exception exception) {
```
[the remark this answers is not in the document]
Sh-Pe: Ok let me clarify what I meant: 
You can catch an exception for two reasons:
- first case: because you know a specific function throws a specific error, and you want to handle that case. If so, you'd catch only that specific error for the reasons written above. 
- The second case: to prevent crashes on runtime. This shouldn't be done for specific lines, but for the code as a whole. And probably should have an option for "debug mode" that disables it. 

You're in the first case. If you're trying to prevent crashes on runtime there's no logic behind taking care of only that specific line.

src/main/java/frc/robot/subsystems/swerve/Swerve.java
```diff
@@ -0,0 +1,297 @@
...
+		this.stateHelper = swerveStateHelper;
+	}
+
+	public void setHeadingSupplier(Supplier<Rotation2d> headingSupplier) {
+		this.headingSupplier = headingSupplier;
+	}
+
+	public void setHeading(Rotation2d heading) {
+		gyro.setYaw(heading);
+		gyro.updateInputs(gyroSignals.yawSignal());
```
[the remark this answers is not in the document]
Sh-Pe: Then run update() first

src/main/java/frc/robot/subsystems/swerve/Swerve.java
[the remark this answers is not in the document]
Sh-Pe: I know they don’t get the gyro. They get null/Rotation2d. In the first case, it breaks annotations and crashes the code. 

What I'm trying to say is that there's not point of having the option to not have a gyro if it crashes the code as a result of miscommunication with other components

src/main/java/frc/robot/subsystems/swerve/Swerve.java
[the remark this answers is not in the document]
Sh-Pe: It means that you have a function called "isatheading" that could be used only for PID controllers

Sh-Pe:
I don't really get the separation between `Swerve` and `Modules`. Swerve is just modules with gyro? And it seems like the separation is not that clear in the code either; the Swerve command builder uses both classes.

src/main/java/frc/robot/subsystems/swerve/Swerve.java
```diff
@@ -0,0 +1,297 @@
+package frc.robot.subsystems.swerve;
```
Sh-Pe, line 1: and empty `.git-blame-ignore-revs` file does not serves an purpose. Note that windows may hide it when you'll try to delete it.

src/main/java/frc/robot/subsystems/swerve/Swerve.java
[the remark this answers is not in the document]
Sh-Pe: like [look at this comment](https://github.com/GreenBlitz/GB-Robot-Template/pull/508#discussion_r1885116359) from 3 weeks ago.
 > Also, I believe this is copied from 6328 and there is no scenario in which the gyro angle is null. If the gyro is disconnected, the swerve breaks anyway

~ (@GilStein1)

When something is not an optional, don't except others to do a null check

Sh-Pe (commented):
Ans

src/main/java/frc/robot/subsystems/swerve/Swerve.java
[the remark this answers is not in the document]
YHerm: is at represent also that you keep in it. not a pid requirement

src/main/java/frc/robot/subsystems/swerve/Swerve.java
emilyyyylime: you can comment on empty files btw

src/main/java/frc/robot/subsystems/swerve/Swerve.java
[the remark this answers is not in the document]
YHerm: i dont agree, i dont see a reason to let the code throw if i can cancel it on the way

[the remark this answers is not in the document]
yonnbro: Don’t be lazy, also, can we assign a value to robot config in the catch block so that the finally statement will not break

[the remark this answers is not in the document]
YHerm: Wdym?

yonnbro (commented):
Added a comment

YHerm:
> I don't really get the separation between `Swerve` and `Modules`. Swerve is just modules with gyro? And it seems like the separation is not that clear in the code either; the Swerve command builder uses both classes.

It’s just for srp and making classes smaller. U r not suppose to use modules

src/main/java/frc/robot/subsystems/swerve/Swerve.java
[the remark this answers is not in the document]
yonnbro: Just specify the exceptions that RobotConfig.fromGUISettings() can throw. And in the catch block assign a value to the robot config var, so that the PP line in finally block will have something to work with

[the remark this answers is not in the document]
hillelv: It won't break, it's value will be the parameter inserted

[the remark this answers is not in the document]
yonnbro: Oh right, correct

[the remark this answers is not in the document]
YHerm: What do you mean the param inserted?
It's null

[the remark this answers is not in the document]
hillelv: the param robotConfig? The third param of the function? Why is it null?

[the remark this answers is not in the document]
YHerm: If the exception was thrown what robot config =?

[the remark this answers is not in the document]
yonnbro: The one passed to the function, I agree it is weird and confusing but I think it works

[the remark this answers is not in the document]
YHerm: oh its param i didnt saw it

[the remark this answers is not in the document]
YHerm: what do you think of getting it out of the function (the try catch), and use only the param?
@hillelv @yonnbro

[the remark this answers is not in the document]
hillelv: You could move the try catch to a pp utils function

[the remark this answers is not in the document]
yonnbro: @tnrz135 you are the mentor to answer that IMO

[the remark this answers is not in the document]
tnrz135: I agree with @hillelv.
This function currently breaks SRP as it tries to get the updated value from the GUI **and** also configure the updated value. From what I understand, this is also part of your issue here, as the part that needs to configured, doesn't need a try/catch block and can simply call `configPathPlanner` with the correct values.

yonnbro (dismissed):

src/main/java/frc/robot/autonomous/AutonomousConstants.java
```diff
@@ -11,4 +14,11 @@ public class AutonomousConstants {
 
 	public static final double CLOSE_TO_TARGET_POSITION_DEADBAND_METERS = 0.5;
 
+	public static final RobotConfig SYNCOPA_ROBOT_CONFIG = new RobotConfig(
+		74,
+		8.6,
+		new ModuleConfig(0.048, 5.24, 0.96, DCMotor.getFalcon500Foc(1), 60, 1),
+		0.577
+	);
+
```
TheOnion313, line 23: All these numbers need to be stored as separate constants, and then applied to the constant RobotConfig

- [ ] move all numbers to separate constants

src/main/java/frc/robot/subsystems/swerve/Swerve.java
```diff
@@ -0,0 +1,286 @@
...
+		this.gyro = gyro;
+		this.gyroSignals = gyroSignals;
+
+		this.kinematics = new SwerveDriveKinematics(modules.getModulePositionsFromCenterMeters());
+		this.headingSupplier = this::getGyroAbsoluteYaw;
+		this.headingStabilizer = new HeadingStabilizer(this.constants);
+		this.stateHandler = new SwerveStateHandler(this);
+		this.commandsBuilder = new SwerveCommandsBuilder(this);
+
+		update();
```
TheOnion313, line 61: Why is this necessary?

src/main/java/frc/robot/subsystems/swerve/Swerve.java
```diff
@@ -0,0 +1,286 @@
...
+	}
+
+	public double getDriveRadiusMeters() {
+		Translation2d[] modulePositionsFromCenterMeters = modules.getModulePositionsFromCenterMeters();
+		double sum = 0;
+		for (Translation2d modulePositionFromCenterMeters : modulePositionsFromCenterMeters) {
+			sum += modulePositionFromCenterMeters.getDistance(new Translation2d());
+		}
+		return sum / modulePositionsFromCenterMeters.length;
+	}
```
TheOnion313, line 161: This whole thing should be a constant calculated at startup

- Change to constant calculated at startup
- change function to return constant / refactor code to use the constant

src/main/java/frc/robot/subsystems/swerve/Swerve.java
```diff
@@ -0,0 +1,286 @@
...
+
+	private void applySpeeds(ChassisSpeeds speeds, SwerveState swerveState) {
+		SwerveModuleState[] swerveModuleStates = kinematics
+			.toSwerveModuleStates(speeds, stateHandler.getRotationAxis(swerveState.getRotateAxis()));
+		setTargetModuleStates(swerveModuleStates, swerveState.getLoopMode().isClosedLoop());
+	}
+
+	private void setTargetModuleStates(SwerveModuleState[] moduleStates, boolean isClosedLoop) {
+		SwerveDriveKinematics.desaturateWheelSpeeds(moduleStates, constants.velocityAt12VoltsMetersPerSecond());
```
TheOnion313, line 271: Why is there a method in the constants class?

- change to constant, or if it is not constant move the method to the proper class

src/main/java/frc/robot/subsystems/swerve/SwerveCommandsBuilder.java
```diff
@@ -0,0 +1,173 @@
...
+		);
+		this.driveCalibrator = new SysIdCalibrator(
+			modules.getModule(ModuleUtils.ModulePosition.FRONT_LEFT).getDriveSysIdConfigInfo(),
+			swerve,
+			modules::setDrivesVoltage
+		);
+	}
+
+	public Command steerCalibration(boolean isQuasistatic, SysIdRoutine.Direction direction) {
+		return steerCalibrator.getSysIdCommand(isQuasistatic, direction).withName("Steer calibration");
```
TheOnion313, line 52: - [ ] store the name of the Command in a constant

src/main/java/frc/robot/subsystems/swerve/SwerveCommandsBuilder.java
```diff
@@ -0,0 +1,173 @@
...
+
+		return new SequentialCommandGroup(
+			pointWheels(new Rotation2d(), false).until(
+				() -> modules.isSteerAtTargetPositions(
+					SwerveConstants.CALIBRATION_MODULE_ANGLE_TOLERANCE,
+					SwerveConstants.CALIBRATION_MODULE_ANGULAR_VELOCITY_PER_SECOND_DEADBAND
+				)
+			),
+			new ParallelDeadlineGroup(sysIdCommand, pointWheels(new Rotation2d(), false))
+		).withName("Drive calibration");
```
TheOnion313, line 67: - [ ] move command name to constant

src/main/java/frc/robot/subsystems/swerve/SwerveCommandsBuilder.java
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+
+	public Command pointWheels(Rotation2d targetSteerPosition, boolean optimize) {
+		return new RunCommand(() -> modules.pointWheels(targetSteerPosition, optimize), swerve)
+			.withName("Point wheels to: " + targetSteerPosition);
```
TheOnion313, line 73: - [ ] move command name to constant

src/main/java/frc/robot/subsystems/swerve/SwerveCommandsBuilder.java
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	public Command pointWheelsInCircle() {
+		return new RunCommand(modules::pointWheelsInCircle, swerve).withName("Point wheels in circle");
```
TheOnion313, line 77: - [ ]  move command name to constant

src/main/java/frc/robot/subsystems/swerve/SwerveCommandsBuilder.java
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	public Command pointWheelsInX() {
+		return new RunCommand(modules::pointWheelsInX, swerve).withName("Point wheels in X");
```
TheOnion313, line 81: - [ ]  move command name to constant

src/main/java/frc/robot/subsystems/swerve/SwerveCommandsBuilder.java
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+
+	public Command setTargetModuleStates(Supplier<SwerveModuleState[]> statesSupplier, boolean isClosedLoop) {
+		return new RunCommand(() -> modules.setTargetStates(statesSupplier.get(), isClosedLoop), swerve).withName("Set states by " + "supplier");
```
TheOnion313, line 86: - [ ]  move command name to constant

src/main/java/frc/robot/subsystems/swerve/SwerveCommandsBuilder.java
```diff
@@ -0,0 +1,173 @@
...
+
+	public Command wheelRadiusCalibration() {
+		return new SequentialCommandGroup(
+			pointWheelsInCircle().until(
+				() -> swerve.getModules()
+					.isSteerAtTargetPositions(
+						SwerveConstants.CALIBRATION_MODULE_ANGLE_TOLERANCE,
+						SwerveConstants.CALIBRATION_MODULE_ANGULAR_VELOCITY_PER_SECOND_DEADBAND
+					)
+			),
```
TheOnion313, line 98: This command can be created in a function, instead of code duplication

- [ ] create command in method and call it everywhere you need it

src/main/java/frc/robot/subsystems/swerve/SwerveCommandsBuilder.java
```diff
@@ -0,0 +1,173 @@
...
+			new WheelRadiusCharacterization(
+				swerve,
+				swerve.getDriveRadiusMeters(),
+				SwerveConstants.WHEEL_RADIUS_CALIBRATION_VELOCITY_PER_SECOND,
+				swerve.getModules()::getDrivesPositions,
+				swerve::getAbsoluteHeading,
+				rotationsPerSecond -> swerve.driveByState(new ChassisSpeeds(0, 0, rotationsPerSecond.getRadians()), SwerveState.DEFAULT_DRIVE),
+				swerve.getModules()::stop
+			)
+		).withName("Wheel radius calibration");
```
TheOnion313, line 108: - [ ]  move command name to constant

src/main/java/frc/robot/subsystems/swerve/SwerveCommandsBuilder.java
```diff
@@ -0,0 +1,173 @@
...
+	public Command turnToHeading(Rotation2d targetHeading) {
+		return turnToHeading(targetHeading, RotateAxis.MIDDLE_OF_CHASSIS);
+	}
+
+	public Command turnToHeading(Rotation2d targetHeading, RotateAxis rotateAxis) {
+		return new InitExecuteCommand(
+			swerve::resetPIDControllers,
+			() -> swerve.turnToHeading(targetHeading, SwerveState.DEFAULT_DRIVE.withRotateAxis(rotateAxis)),
+			swerve
+		).withName("Rotate around " + rotateAxis.name() + " to " + targetHeading);
```
TheOnion313, line 121: - [ ]  move command name to constant (and use format string)

src/main/java/frc/robot/subsystems/swerve/SwerveCommandsBuilder.java
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+
+	public Command drive(DoubleSupplier xSupplier, DoubleSupplier ySupplier, DoubleSupplier rotationSupplier) {
+		return driveByState(xSupplier, ySupplier, rotationSupplier, SwerveState.DEFAULT_DRIVE).withName("Default drive");
```
TheOnion313, line 126: - [ ]  move command name to constant

src/main/java/frc/robot/subsystems/swerve/SwerveCommandsBuilder.java
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	public Command driveByState(
+		DoubleSupplier xSupplier,
+		DoubleSupplier ySupplier,
+		DoubleSupplier rotationSupplier,
+		Supplier<SwerveState> state
+	) {
+		return new DeferredCommand(() -> driveByState(xSupplier, ySupplier, rotationSupplier, state.get()), Set.of(swerve))
+			.withName("Drive with supplier state");
```
TheOnion313, line 136: I'm gonna stop, change all the in place strings to constants

TheOnion313 (changes requested):

src/main/java/frc/robot/subsystems/swerve/Swerve.java
YHerm: This method update all getters and logs of the swerve. When I create swerve I would want all of this info to be updated

src/main/java/frc/robot/subsystems/swerve/Swerve.java
YHerm: It’s record. The swerve constants file is a record of constants. Each swerve can have different constants but these values don’t change through the iteration

drorta (approved):

src/main/java/frc/robot/subsystems/swerve/states/SwerveStateHandler.java
```diff
@@ -0,0 +1,82 @@
...
+
+	public void setObjectTranslationSupplier(Supplier<Optional<Translation2d>> objectTranslationSupplier) {
+		this.objectTranslationSupplier = objectTranslationSupplier;
+	}
+
+	public ChassisSpeeds applyAimAssistOnChassisSpeeds(ChassisSpeeds speeds, SwerveState swerveState) {
+		if (swerveState.getAimAssist() == AimAssist.NONE) {
+			return speeds;
+		}
+		return speeds;
```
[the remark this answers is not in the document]
YHerm: in future there will be more options so it will look less weird.
now for why the if and not none as default? most of the times the state will be NONE so putting it first in the if order will give u a better runtime

yonnbro (approved):

src/main/java/frc/robot/subsystems/swerve/states/SwerveStateHandler.java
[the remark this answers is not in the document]
YHerm: @Nitay4

Nitay4 (approved):

PR #565, opened 2025-01-04
Gui Config Improve -> core-swerve

YHerm:


src/main/java/frc/utils/auto/PathPlannerUtils.java
```diff
@@ -58,6 +47,23 @@ public static void scheduleWarmup() {
...
+
+	private static void reportAlert(Alert.AlertType alertType, String message) {
+		new Alert(alertType, AutonomousConstants.LOG_PATH_PREFIX + message).report();
+	}
+
+	public static Optional<RobotConfig> getGuiRobotConfig() {
+		try {
+			RobotConfig robotConfig = RobotConfig.fromGUISettings();
+			return Optional.of(robotConfig);
+		} catch (IOException ioException) {
```
hillelv, line 59: If you want to be specific it also trows FileNotFoundException

YHerm: image

hillelv: image

YHerm: the fun throws IO Exception so u must catch IO Exception 
image


i can catch IO and print file not found if u want
@hillelv

hillelv: It trows both inside, file not found extends io

PR #566, opened 2025-01-05
updated pplib vendor -> master

YHerm:


hillelv (dismissed):

drorta (dismissed):

hillelv:
navgrid tells the pathfinding where the field obstacles are. if we put it in gitignore it can cause a bug where we avoid non existent obstacles or, more importantly, not avoid existing obstacles

Sh-Pe (approved):

Sh-Pe:
PP `seems to cause compile errors](https://github.com/GreenBlitz/GB-Robot-Template/actions/runs/12725813790/job/35473474096?pr=501), does PR has anything to do with fixing it? 
and pls don't break master again 😅

hillelv:
Yes it does, and pp doesn't break master 🙈

PR #567, opened 2025-01-05
Romy & Noam - added the filter classes to odometry

NoamMilgrom:


src/main/java/frc/robot/odometry/OdometryFilters.java
```diff
@@ -0,0 +1,5 @@
+package frc.robot.odometry;
+
+public class OdometryFilters {
+
```
NoamRosenberg08, line 4: Some of the filters need to get the data from listeners, like acceleration or values of accelerometers. Consider adding. A utils class for that. 

Maybe for another branch

NoamRosenberg08 (changes requested):

NoamRosenberg08 (approved):

RJ0907 (approved):

src/main/java/frc/robot/poseestimator/observations/OdometryFilterer.java
```diff
@@ -0,0 +1,21 @@
...
+
+    private final Filter<OdometryObservation> filter;
+
+    public OdometryFilterer(Filter<OdometryObservation> filter) {
+        this.filter = filter;
+    }
+
+    public Optional<OdometryObservation> getFilteredObservation(OdometryObservation observation) {
+        if (!filter.apply(observation)) {
+            observation = null;
```
hillelv, line 17:
```suggestion
            return Optional.empty();
```

src/main/java/frc/robot/poseestimator/observations/OdometryFilterer.java
```diff
@@ -0,0 +1,21 @@
+package frc.robot.poseestimator.observations;
+
+import frc.utils.Filter;
+
+import java.util.Optional;
+
+public class OdometryFilterer {
+
+    private final Filter<OdometryObservation> filter;
```
hillelv, line 9: Consider making the filter changeable so we won't need to create a new one each time we change the filter

TheOnion313 (dismissed):

src/main/java/frc/robot/odometry/OdometryFilterer.java
```diff
@@ -0,0 +1,28 @@
...
+public class OdometryFilterer {
+
+    private Filter<OdometryObservation> filter;
+
+    public OdometryFilterer(Filter<OdometryObservation> filter) {
+        this.filter = filter;
+    }
+
+    public void addFilter(Filter<OdometryObservation> filter) {
+        this.filter.and(filter);
```
hillelv, line 17: what if I want or?

src/main/java/frc/robot/odometry/OdometryFilterer.java
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    public void clearFilter() {
+        this.filter = new Filter<>(data -> true);
+    }
+
+    public Optional<OdometryObservation> getFilteredObservation(OdometryObservation observation) {
+        return !filter.apply(observation) ? Optional.empty() : Optional.of(observation);
```
hillelv, line 25:
```suggestion
        return filter.apply(observation) ? Optional.of(observation) : Optional.empty();
```

src/main/java/frc/robot/odometry/OdometryFilterer.java
```diff
@@ -0,0 +1,28 @@
+package frc.robot.odometry;
+
+import frc.robot.poseestimator.observations.OdometryObservation;
+import frc.utils.Filter;
+
+import java.util.Optional;
+
+public class OdometryFilterer {
```
Sh-Pe, line 8: Why don't you extend the Filter class? It will save you most of the code here

src/main/java/frc/robot/odometry/OdometryFilters.java
```diff
@@ -0,0 +1,5 @@
+package frc.robot.odometry;
+
+public class OdometryFilters {
```
Sh-Pe, line 3: If you have time, check filtering constant velocity reported by the swerve. Usually happens when the robot is not actually moving. I don't know if it'll work though. 
Also, odometryObservation is not robust enough for most of the filters. You need the prev observation to calc the twist, and the observation by its own is not really useful. And for that you need the kinematics too. 

I think a more dynamic, odometryFilterer object, that is non static and has a constructor, that can get the kinematics and calculated the twist would be much more useful. Those filters could also fit the Filter class, assuming you find a way to define it by a function that takes more than just odometryObservation. 
ofc, there is a place for static filters, like for acceleration filtering (though this PR is kinda stuck right now, no one really does PR in the season), if herman would had acceleration to odometryObservation.

Sh-Pe:
draft/closed?

PR #582, opened 2025-01-17
Filters Manager ⤳ PoseEstimator-Master

Sh-Pe:
## Linear filters but wrap, to reduce repetitive usage of linear filter within the code. 

So what is a filter? A filter is an object that takes input over times, and returns a modified version of it. It differs from a supplier since because the filter is an object (and not an interface) so it's able to store previous data (usually in using a RingBuffer) and preform the calculation based on it. 

### Use cases: 
- Denosing pigeon's output (acceleration & angular velocity measures).
- Denosing vision.
- Using advanced averages simply
Currently, the first one isn't merged yet, and for the second one, we would like to have a poseestimator merged before adding filters on it. 

### Current uses
This PR should set the ground for future uses in this season. 
idk really how to set labels, it's not really poseestimator nor robot. 

### other
- **About *linear* filters:** [WPILib's site](https://docs.wpilib.org/en/stable/docs/software/advanced-controls/filters/linear-filter.html)


src/main/java/frc/utils/linearfilters/Periodic3DlLinearFilter.java
```diff
@@ -0,0 +1,62 @@
...
+	public void hardReset() {
+		linearFilters.forEach(LinearFilter::reset);
+	}
+
+	@Override
+	public void log(String parentLogPath) {
+		String logPath = parentLogPath + name + "/";
+		for (int i = 0; i < updateValues.get().size(); i++) {
+			Logger.recordOutput(logPath + "input/" + i, updateValues.get().get(i));
+		}
```
DANNA2828, line 41:
```suggestion
	public void log(String parentLogPath) {
		String inputsLogPath = parentLogPath + name + "/Inputs/";
		for (int i = 0; i < updateValues.get().size(); i++) {
			Logger.recordOutput(inputsLogPath + i, updateValues.get().get(i));
		}
```

src/main/java/frc/utils/linearfilters/PeriodicLinearFilter.java
```diff
@@ -0,0 +1,46 @@
...
+
+	public double getOutput() {
+		return filter.lastValue();
+	}
+
+	public void hardReset() {
+		filter.reset();
+	}
+
+	public void log(String parentLogPath) {
```
DANNA2828, line 32: same as comment on other file

src/main/java/frc/utils/linearfilters/Periodic3DlLinearFilter.java
```diff
@@ -0,0 +1,62 @@
...
+		Logger.recordOutput(logPath + "output", getAsColumnVector());
+	}
+
+	@Override
+	public void update() {
+		for (Vector<N3> data : updateValues.get()) {
+			for (int i = 0; i < 3; i++) {
```
DANNA2828, line 48: magic number
do linear filters.leangth

DANNA2828 (commented):
hihihiha

src/main/java/frc/utils/linearfilters/Periodic3DlLinearFilter.java
```diff
@@ -0,0 +1,62 @@
...
+			}
+		}
+	}
+
+	public Vector<N3> getAsColumnVector() {
+		return new Vector<>(new SimpleMatrix(new double[][] {getAsPoseArray()}));
+	}
+
+	private double[] getAsPoseArray() {
+		return new double[] {linearFilters.get(0).lastValue(), linearFilters.get(1).lastValue(), linearFilters.get(2).lastValue()};
```
GilStein1, line 59:
```suggestion
		double[] poseArray = new double[linearFilters.size()];
		for(int i = 0; i < linearFilters.size(); i++) {
		    poseArray[i] = linearFilters.get(i).lastValue();
		}
		return poseArray;
```

GilStein1 (changes requested):
Maybe instead of the manager having the suppliers in it, make a wrapper for each filter that holds the supplier instead

src/main/java/frc/utils/linearfilters/Periodic3DlLinearFilter.java
Sh-Pe: It's much longer and does no serve any purpose

GilStein1: it is better than having magic numbers

src/main/java/frc/utils/linearfilters/Periodic3DlLinearFilter.java
Sh-Pe: Not used only here

src/main/java/frc/utils/linearfilters/Periodic3DlLinearFilter.java
Sh-Pe: I've solved this with streams, can I resolve?

src/main/java/frc/utils/filters/FiltersManager.java
```diff
@@ -0,0 +1,52 @@
...
+	public static void addFilter(IPeriodicFilter<?> filter) {
+		periodicLinearFilters.add(filter);
+	}
+
+	/**
+	 * for filters that uses past measurements for filtering, reset the past data.
+	 */
+	public static void resetAllFilters() {
+		periodicLinearFilters.forEach(IPeriodicFilter::hardReset);
+	}
```
Sh-Pe, line 34: why

src/main/java/frc/utils/filters/linearfilters/PeriodicNDimlLinearFilter.java
```diff
@@ -0,0 +1,74 @@
...
+
+	/**
+	 * creates a n-dim periodic linear filter
+	 *
+	 * @param updateValues: supplies all the values (as n-dim vectors) in the current periodic circle
+	 * @param filters:      n-sized list, that injects the filters for each value of the given vector
+	 * @param name:         the name of the periodic filter
+	 * @param sizeInstance: an instance of wpilib's Num (represent the size of the vectors)
+	 */
+	public PeriodicNDimlLinearFilter(Supplier<List<Vector<T>>> updateValues, List<LinearFilter> filters, String name, T sizeInstance) {
```
Sh-Pe, line 30: List is too restrictive

src/main/java/frc/utils/filters/linearfilters/PeriodicNDimlLinearFilter.java
```diff
@@ -0,0 +1,74 @@
...
+	public void log(String parentLogPath) {
+		String logPath = parentLogPath + name + "/";
+		for (int i = 0; i < updateValues.get().size(); i++) {
+			for (int j = 0; j < size.getNum(); j++) {
+				Logger.recordOutput(logPath + "lastUpdate" + i + "/" + j, TimeUtils.getCurrentTimeSeconds());
+				Logger.recordOutput(logPath + "input" + i + "/" + j, updateValues.get().get(i).get(0, j));
+			}
+		}
+		for (int i = 0; i < size.getNum(); i++) {
+			Logger.recordOutput(logPath + "output/" + i, get().get(0, i));
```
Sh-Pe, line 53: you don't need that many logs

PR #584, opened 2025-01-19
New swerve -> master

DANNA2828:


src/main/java/frc/JoysticksBindings.java
```diff
@@ -153,6 +153,8 @@ private static void secondJoystickButtons(Robot robot) {
 	private static void thirdJoystickButtons(Robot robot) {
 		SmartJoystick usedJoystick = THIRD_JOYSTICK;
 		// bindings...
+
+		usedJoystick.B.onTrue(robot.getSwerve().getCommandsBuilder().driveCalibration());
```
Nitay4, line 157: are you sure this is supposed to be here

src/main/java/frc/robot/subsystems/swerve/factories/swerveconstants/RealSwerveConstants.java
```diff
@@ -6,11 +6,11 @@
 
 public class RealSwerveConstants {
 
-	public static final double VELOCITY_AT_12_VOLTS_METERS_PER_SECOND = 5.054;
+	public static final double VELOCITY_AT_12_VOLTS_METERS_PER_SECOND = 3.8;//5.054;
```
Nitay4, line 9: why the comment?

hillelv:
I think you should make different files for the new chassis and syncopa it would be annoying to move to different branches or repos to run different robots

PR #631, opened 2025-02-06
Yaw angle filter -> PE master

Yotam7:


src/main/java/frc/robot/vision/VisionFilters.java
```diff
@@ -8,6 +8,8 @@
 import frc.utils.Filter;
 import frc.utils.math.ToleranceMath;
 
+import java.util.function.Supplier;
+
 public class VisionFilters {
```
GilStein1, line 13: Put the angle filters in the XYZ order.
First the roll, then the pitch and then the yaw

DANNA2828 (approved):

GilStein1 (approved):

hillelv (approved):

PR #632, opened 2025-02-08
PoseEstimator bugfixes -> poseestimator-master

GilStein1:


src/main/java/frc/robot/vision/sources/limelights/LimeLightSource.java
```diff
@@ -108,8 +110,7 @@ protected double getLatency() {
 	}
 
 	protected Optional<Pair<Pose3d, Double>> getUpdatedPose3DEstimation() {
-		int id = getAprilTagID();
-		if (id == VisionConstants.NO_APRILTAG_ID) {
+		if (lastSeenAprilTagId == VisionConstants.NO_APRILTAG_ID) {
 			return Optional.empty();
 		}
```
Sh-Pe, line 115: There is that network table entry to get that info (whether a target is seen or not). Useful in case we would want to use a dual target mode.

src/main/java/frc/utils/math/DataMath.java
```diff
@@ -0,0 +1,22 @@
+package frc.utils.math;
+
+import java.util.function.Function;
+
+public class DataMath {
+
+	public static <T extends Iterable<E>, E> double calculateStandardDeviations(T values, Function<E, Double> getValue) {
```
Sh-Pe, line 7:
```suggestion
	public static <T extends Iterable<E>, E> double calculateStandardDeviations(T iterable, Function<E, Double> getValue) {
```

src/main/java/frc/utils/math/DataMath.java
```diff
@@ -0,0 +1,22 @@
...
+		int length = 0;
+		double total = 0;
+		for (E value : values) {
+			length++;
+			total += getValue.apply(value);
+		}
```
Sh-Pe, line 13: Use streams (one short like instead of 5). Consider composing with .limit to avoid getting into an infinite loop (from a defensive programming perspective).

src/main/java/frc/utils/TriFunction.java
```diff
@@ -0,0 +1,7 @@
+package frc.utils;
+
+public interface TriFunction<In1, In2, In3, Out> {
+
+	Out apply(In1 var1, In2 var2, In3 var3);
+
```
Sh-Pe, line 6:
```suggestion
static class TriValueFuncion<In, Out> extend TriFunction<In, In, In, Out> {}
```
Will save you space every time you need the same 3 vals. Not must.

Sh-Pe (approved):

src/main/java/frc/utils/math/DataMath.java
yonnbro, line 1: This is a terrible name

src/main/java/frc/utils/TriFunction.java
yonnbro: Typo in @Sh-Pe comment

src/main/java/frc/utils/TriFunction.java
```diff
@@ -0,0 +1,7 @@
+package frc.utils;
+
+public interface TriFunction<In1, In2, In3, Out> {
```
yonnbro, line 3: Add a short description of what this is and an example to look at a usage of this interface

yonnbro (approved):

yonnbro (approved):

PR #630, opened 2025-02-04
Poseestimator Release 2.0 -> master

GilStein1:
⣿⣿⣿⣿⣿⣿⣿⡿⠛⠉⠁⢀⠈⠉⠻⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿
⣿⣿⣿⣿⣿⣿⠋⢀⣴⣾⡟⠈⠸⣷⣄⠈⢻⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿
⣿⣿⣿⣿⣿⣿⢀⣾⠉⣿⡇⠀⢀⣴⣿⣆⠘⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿
⣿⣿⣿⣿⣿⡏⣀⢁⡄⢹⣿⣷⣾⣿⣿⣿⡀⢸⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿
⣿⣿⣿⣿⠟⣠⣛⣨⣷⠀⠛⣿⣿⡿⠇⣿⡇⢸⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿
⣿⡿⠛⣡⣾⣿⡿⠟⠉⠂⠀⠉⣁⣤⣾⣿⠇⢸⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿
[...]

PR #648, opened 2025-04-09
Vision Changes From the Season to PoseEstimator-Master (pending)

Sh-Pe:
### This PR is Pending for PPv2.0 to be merged in order to get into Master. Please consider reviewing https://github.com/GreenBlitz/GB-Robot-Template/pull/630 before reviewing this PR

This PR includes the following changes: 
- Filter<T> as a functional interface
- More VisionFilters (thanks to @GilStein1)
- Limelights now send inside its VisionData the PoseEstimating method
- VisionData includes the source's name
- [not from the season] log() is now part of the VisionSource interface, and MultiVisionSources delegates it instead of preforming its own logging. This means sources does not automatically logs its data, and the runtime cost of logging could be measured separately of general source updates

The following changes from the season are *not* included in this PR and still waiting to get merged: 
- HeadingEstimator improvements
- (optional) some constants, including the default vision filter
- improved OdometryData performance (thank to @YHerm)
- Other PP changes

### DIFF
--- a/src/main/java/frc/constants/VisionConstants.java
+++ b/src/main/java/frc/constants/VisionConstants.java
@@ -10,10 +10,6 @@ import java.io.IOException;
 
 public class VisionConstants {
 
-	public static final String FILTERED_DATA_LOGPATH_ADDITION = "FilteredData/";
-
-	public static final String NON_FILTERED_DATA_LOGPATH_ADDITION = "NonFilteredData/";
-
 	public static final String VISION_SOURCE_LOGPATH_ADDITION = "VisionSource/";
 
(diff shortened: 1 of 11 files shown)


YHerm (commented):
i think this PR is hugeeeeee. Consider to split it (things will get merged A LOT faster)

Sh-Pe:
@YHerm these are 175 lines of code. I don’t see any point in that. With that said I can split it if it's needed. 

@GilStein1 wants this in release 2 which is already (really, this time) huge. 
So it seems like this PR is even pending.

PR #651, opened 2025-04-09
Filters From season to PE master

Sh-Pe:
changelog:
- Filter<T> is now a functional interface
- Filters for specifically MT1/2
not pending. Will be part of PPv2.0


Sh-Pe:
copy paste from dumbo code. however I would want to test on the robot just to be sure

src/main/java/frc/robot/vision/VisionFilters.java
```diff
@@ -2,65 +2,95 @@
...
+
+	public static Filter<AprilTagVisionData> isNotSeeingTags(int... tags) {
+		Filter<AprilTagVisionData> filter = isSeeingTag(VisionConstants.NO_APRILTAG_ID);
+
+		for (int id : tags) {
+			filter = filter.or(isSeeingTag(id));
+		}
+
+		return filter.not();
+	}
```
DANNA2828, line 94: This filters out the data if you are not seeing any tags which is not what the function tries to do
Or at least change the name

src/main/java/frc/robot/vision/VisionFilters.java
```diff
@@ -4,63 +4,87 @@
...
+			.orElse(false);
+	}
+
+	public static Filter<VisionData> isYawAtAngleForMegaTag2(Supplier<Optional<Rotation2d>> wantedYawSupplier, Rotation2d yawTolerance) {
+		return isDataFromMegaTag2().implies(isYawAtAngle(wantedYawSupplier, yawTolerance));
+	}
+
+	/**
+	 * A Filter that filters the random noise in MegaTag2 that causes the yaw angle to be exactly 0.
+	 */
```
GilStein1, line 49: The noise does cause the yaw angle to be exactly 0, but in some rare instances, it causes it to be around 0.
I don't think that there is a reason to put this comment here

src/main/java/frc/robot/vision/VisionFilters.java
```diff
@@ -4,63 +4,87 @@
...
 		return isXInField(positionToleranceMeters).and(isYInField(positionToleranceMeters));
 	}
 
+	public static Filter<AprilTagVisionData> isSeeingTag(int tagID) {
+		return (visionData) -> visionData.getTrackedAprilTagId() == tagID;
+	}
+
+	public static Filter<AprilTagVisionData> isNotSeeingTags(int... tags) {
+		return Filter.orAll(Arrays.stream(tags).mapToObj(VisionFilters::isSeeingTag).toList()).not();
+	}
```
GilStein1, line 88: Oy Vey. This code might not be readable for some people, consider using other approaches that are easier to understand

src/main/java/frc/robot/vision/data/LimeLightAprilTagVisionData.java
```diff
@@ -0,0 +1,29 @@
...
+	) {
+		super(sourceName, estimatedRobotPosition, timestamp, standardDeviations, aprilTagHeightMeters, distanceFromAprilTagMeters, aprilTagId);
+		this.poseEstimationMethod = poseEstimationMethod;
+	}
+
+	public LimelightPoseEstimationMethod getPoseEstimationMethod() {
+		return poseEstimationMethod;
+	}
+
+}
```
GilStein1, line 29: A bit out of scope but i don't really care, you can go ahead and resolve this one

GilStein1 (changes requested):

src/main/java/frc/robot/vision/data/LimeLightAprilTagVisionData.java
Sh-Pe: it's used for the new filters but I get what you're saying. idc since no one else is going to review this anyway

src/main/java/frc/robot/vision/VisionFilters.java
Sh-Pe: what's the problem with that? It takes each tag and maps it to isSeeingTag(tag), so you end up with a list of Filters. 
It's not some kind of really nested stream, just mapping

src/main/java/frc/robot/vision/VisionFilters.java
Sh-Pe: I think you wrote this comment in the first place but it doesn't really matter anyways

src/main/java/frc/robot/vision/VisionFilters.java
Sh-Pe: May you use more Hebrew since I can't understand that /s
I get your point but this could be said for many other features of Java that got merged into the codebase (e.g. final functions, functional interfaces, threads, etc.)

GilStein1: Ok, it is already in our code and at this point it will be a great תמריץ to learn about streams for those who want to anderstand the code. I am resolving this one

GilStein1 (approved):

src/main/java/frc/robot/vision/VisionFilters.java
Yotam7: I tend to agree with @Sh-Pe on this issue. I don't think we should ban useful java features just because people aren't familiar with them.
If we use streams (or anything else), we can just include them as part of הכשרות.
What percentage of team members do you think can understand these lines?
https://github.com/GreenBlitz/GB-Robot-Template/blob/34156b94a47e92f9cff759548ece065a2e765328/src/main/java/frc/robot/vision/sources/VisionSource.java#L31
https://github.com/GreenBlitz/GB-Robot-Template/blob/34156b94a47e92f9cff759548ece065a2e765328/src/main/java/frc/utils/math/StatisticsMath.java#L7

PR #652, opened 2025-04-09
VisionData optimizations on runtime time to [pending] Master

Sh-Pe:
Does not impact how to code works, just makes the jvm bytecode to run faster

pending for #651


YHerm (commented):
I dont think we should add this now. For now the convention is not doing final on funcs. maybe in future we will start adding final but its not for now and should be added as convention to all funcs if adding it

GilStein1 (approved):

PR #653, opened 2025-04-09
vision logging bugfix to master

Sh-Pe:
useless because costly on runtime. but may be used for debugging in the future, and it anyway was on dumbo.


Sh-Pe:
actual no need robot testing

GilStein1 (approved):

PR #655, opened 2025-04-10
[Refactor] frc/constants/VisionConstants to vision/VisionConstants

Sh-Pe:
Because Herman asked 

It was like that on Reefscape.


Sh-Pe:
no need to test a refactor

YHerm (approved):

PR #672, opened 2025-04-25
Herman's CR to PE master

Sh-Pe:
I guess I have to open a PR for that because I shouldn't just make commits into PE-master

have fun

refactors only so no need to test


Sh-Pe:
@GilStein1 approve pls

GilStein1 (approved):

Sh-Pe:
Can I merge?

PR #686, opened 2025-05-08
Move to object -> master

Nitay4:


src/main/java/frc/robot/subsystems/swerve/SwerveCommandsBuilder.java
```diff
@@ -214,4 +226,35 @@ public Command moveToPoseByPID(Supplier<Pose2d> currentPose, Pose2d targetPose)
 		);
 	}
 
+	public Command moveToPoseByPID(Supplier<Pose2d> currentPose, Supplier<Pose2d> targetPose) {
+		return swerve.asSubsystemCommand(
+			new InitExecuteCommand(swerve::resetPIDControllers, () -> swerve.moveToPoseByPID(currentPose.get(), targetPose.get())),
+			"PID to pose: " + targetPose
+		);
+	}
```
hillelv, line 234: aint no way you copied the previous func instead of changing it

src/main/java/frc/robot/subsystems/swerve/SwerveCommandsBuilder.java
```diff
@@ -214,4 +226,35 @@ public Command moveToPoseByPID(Supplier<Pose2d> currentPose, Pose2d targetPose)
...
+
+	public Command findObject(Supplier<Optional<Translation2d>> objectTranslation) {
+		return swerve.asSubsystemCommand(spin().until(objectTranslation.get()::isPresent), "Find object");
+	}
```
hillelv, line 238: I think this shouldn't be a function. this relies on the fact that the object detection returns an optional of a translation which could change depending on the implementation. if you just get a boolean supplier then it just becomes a wrapper for until which is unnecessary

src/main/java/frc/robot/subsystems/swerve/SwerveCommandsBuilder.java
```diff
@@ -214,4 +226,35 @@ public Command moveToPoseByPID(Supplier<Pose2d> currentPose, Pose2d targetPose)
...
+
+	public Command driveToObject(Supplier<Pose2d> currentPose, Supplier<Optional<Translation2d>> objectTranslation, double distance) {
```
hillelv, line 240: distance of what?

src/main/java/frc/robot/subsystems/swerve/SwerveCommandsBuilder.java
```diff
@@ -214,4 +226,35 @@ public Command moveToPoseByPID(Supplier<Pose2d> currentPose, Pose2d targetPose)
...
+		return swerve.asSubsystemCommand(
+			new SequentialCommandGroup(
+				findObject(objectTranslation).until(() -> objectTranslation.get().isPresent()),
```
hillelv, line 243: bruh did you just put until on the function that only exists because there is an until inside 💀

src/main/java/frc/robot/subsystems/swerve/SwerveCommandsBuilder.java
```diff
@@ -214,4 +226,35 @@ public Command moveToPoseByPID(Supplier<Pose2d> currentPose, Pose2d targetPose)
...
+				moveToPoseByPID(
+					currentPose,
+					() -> objectTranslation.get().isPresent()
+						? FieldMath.findPointByDistanceFromAnotherPointAndAngleFromASecondPoint(
+							objectTranslation.get().get(),
+							currentPose.get(),
+							distance
+						)
+						: currentPose.get()
```
hillelv, line 252: optional has this built in but I don't remember the sintax

src/main/java/frc/utils/math/FieldMath.java
```diff
@@ -71,4 +71,9 @@ public static Translation2d mirror(Translation2d translation2d, boolean mirrorX,
 		return translation2d;
 	}
 
+	public static Pose2d findPointByDistanceFromAnotherPointAndAngleFromASecondPoint(Translation2d pointA, Pose2d pointB, double distance) {
```
hillelv, line 74: מה לעזאסיגמה

PR #813, opened 2026-07-30
imu signals

Dolev3:


src/main/java/frc/robot/subsystems/swerve/Swerve.java
```diff
@@ -97,17 +97,16 @@ public SwerveStateHandler getStateHandler() {
...
 
 	public Rotation3d getOrientationFromIMU() {
-		return imuSignals.getOrientation();
+		return imuSignals.getLatestOrientation();
 	}
 
 	public Translation3d getAccelerationFromIMUMetersPerSecondSquared() {
-		return imuSignals.getAccelerationEarthGravitationalAcceleration()
-			.times(RobotConstants.GRAVITATIONAL_ACCELERATION_METERS_PER_SECOND_SQUARED_ISRAEL);
+		return imuSignals.getLatestAccelerationG().times(RobotConstants.GRAVITATIONAL_ACCELERATION_METERS_PER_SECOND_SQUARED_ISRAEL);
```
spiderh0ok: G

spiderh0ok (dismissed):

YHerm (dismissed):

src/main/java/frc/robot/subsystems/swerve/Swerve.java
```diff
@@ -97,17 +97,16 @@ public SwerveStateHandler getStateHandler() {
 		return stateHandler;
 	}
 
-	public Rotation3d getAngularVelocityFromIMURotation2dPerSecond() {
-		return imuSignals.getAngularVelocity();
+	public Rotation2d[] getIMUAngularVelocityRPS() {
```
hillelv, line 100: RPS?

src/main/java/frc/robot/subsystems/swerve/Swerve.java
```diff
@@ -234,7 +233,7 @@ private ChassisSpeeds getDriveModeRelativeSpeeds(ChassisSpeeds speeds, SwerveSta
 	}
 
 	public double getIMUAcceleration() {
```
hillelv: Units?

src/main/java/frc/robot/subsystems/swerve/Swerve.java
spiderh0ok: I think its ok

hillelv: Is it revolutions per second, radians per second, rotation2d per second?

src/main/java/frc/robot/subsystems/swerve/IMUSignals.java
```diff
@@ -10,32 +10,56 @@ public record IMUSignals(
...
-			this.pitchSignal().getLatestValue().getRadians(),
-			this.yawSignal().getLatestValue().getRadians()
+			rollSignal.getLatestValue().getRadians(),
+			pitchSignal.getLatestValue().getRadians(),
+			yawSignal.getLatestValue().getRadians()
 		);
 	}
 
-	public Translation3d getAccelerationEarthGravitationalAcceleration() {
+	public Translation3d[] getAllAccelerationsG() {
```
Sh-Pe: Not a clear name. We have acceleration is 6 different axes (rotation + position), it should be clear your talking about the second derivative of the robot's position in the 3D space (and not the orientation).

src/main/java/frc/robot/subsystems/swerve/Swerve.java
Sh-Pe: The function returns Rotation2d, so it can be only the latter

Sh-Pe:
Just wanted to comment here that previously you argued with me for like a week that it's not clear enough that "G" is the gravitational constant. 

Also it's worth checking whether the IMU output is actually dependent on the place on earth (nor this fix g=9.8 or smth) since the constant changes when we go to Houston. 

And it should be $g$ and not $G$, since $G$ is the global gravitational constant.

src/main/java/frc/robot/subsystems/swerve/Swerve.java
Dolev3: herman aggred that it should stay RPS as in Rotation2D per second

spiderh0ok (dismissed):

hillelv (dismissed):

PR #825, opened 2026-08-01
Is continuous wrap testing -> rebuilt tester

rys4590:


src/main/java/frc/robot/subsystems/HoodConstants.java
```diff
@@ -17,6 +17,7 @@
 public class HoodConstants {
 
     public static final boolean IS_INVERTED = true;
+    public static final boolean IS_CONTINUOUS_WRAP = false;
```
spiderh0ok: for the test make it true, i think we will increase the gear ratio to test if it works, so it thinks it did a full circle

